Issue #22, opened 2018-02-05
Proposal: Date support


4F2E4A2E:
I propose to add Date support using this [implementation](https://gist.github.com/langpavel/b30f3d507a47713b0c6e89016e4e9eb7), found here https://github.com/graphql/graphql-js/issues/550

19majkel94:
Honestly, I prefer the ISO version of the Date scala, it's human readable and easier to debug:
```ts
import { GraphQLScalarType, Kind } from "graphql";

export const DateScalar = new GraphQLScalarType({
  name: "Date",
  description: "Date scalar type",
  parseValue(value: string) {
    return new Date(value); // value from the client input variables
  },
  serialize(value: Date) {
    return value.toISOString(); // value sent to the client
  },
  parseLiteral(ast) {
    if (ast.kind === Kind.STRING) {
      return new Date(ast.value); // value from the client query
    }
    return null;
  },
});
```
And because I don't want to make this framework so opinionated, I prefer to gave developers freedom to  use their own version of `Date` and other scalars with #5 custom scalar support 😉

4F2E4A2E:
You are right.  I will change it to the iso format.

4F2E4A2E:
A default date speeds up the getting started, but of course, passing in the custom scalar type as presented by you is just fine too:
```typescript
import { GraphQLScalarType, Kind } from 'graphql';
import { Field, GraphQLObjectType } from 'type-graphql';

export const DateScalar = new GraphQLScalarType({
  name:        'Date',
  description: 'Date scalar type',
  parseValue(value: string) {
    return new Date(value); // value from the client input variables
  },
  serialize(value: Date) {
    return value.toISOString(); // value sent to the client
  },
  parseLiteral(ast) {
    if (ast.kind === Kind.STRING) {
      return new Date(ast.value); // value from the client query
    }
    return null;
  }
});

@GraphQLObjectType()
export class User {
  @Field(type => DateScalar)
  contributorsince: Date;
  @Field()
  alive: boolean;
}
```

19majkel94:
Closing - implemented in ba21848 😉

PR #23, opened 2018-02-05
Proposal: #22 default date support

4F2E4A2E:
Adding Date support

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,13 @@ node_modules
 .vscode
 build
 package
+.idea/
+tests/**/*.js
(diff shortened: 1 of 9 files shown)


19majkel94:
@4F2E4A2E 
Thank you very much for your contribution! 😉 
I've implemented `Date` support for both scalars version - ISODate is default but you can change it with `dateScalarMode` option of `buildSchema`.
You can even overwrite this with you own scalar types map with `scalarsMap` option, it's great e.g. for Mongo's ObjectId.

Closing this PR - ba21848 has implemented it.

PR #25, opened 2018-02-07
Small project setup improvements

4F2E4A2E:
Please don't mind the removal of unused imports: if it is something, that is needed at runtime, then i propose to explicitly document it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,10 @@ Currently released version is an early alpha. However it's working quite well, s
 
 More feedback = less bugs thanks to you! :smiley:
 
+## Contribution
+PRs are welcome, but first check, test and build your code before committing it and use commit rules.
+For more information checkout this [commit rule guide](https://gist.github.com/stephenparish/9941e89d80e2bc58a153).
+
 ## Roadmap
 You can keep track of [development's progress on project board](https://github.com/19majkel94/type-graphql/projects/1).
 
--- a/README.md
+++ b/README.md
@@ -380,8 +380,9 @@ Currently released version is an early alpha. However it's working quite well, s
 More feedback = less bugs thanks to you! :smiley:
 
 ## Contribution
-PRs are welcome, but first check, test and build your code before committing it and use commit rules.
-For more information checkout this [commit rule guide](https://gist.github.com/stephenparish/9941e89d80e2bc58a153).
+PRs are welcome, but first check, test and build your code before committing it.
+* Use commit rules: For more information checkout this [commit rule guide](https://gist.github.com/stephenparish/9941e89d80e2bc58a153).
+* [Allowing changes to a pull request branch created from a fork](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)
 
 ## Roadmap
 You can keep track of [development's progress on project board](https://github.com/19majkel94/type-graphql/projects/1).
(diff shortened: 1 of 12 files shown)


src/scalars/aliases.ts
```diff
@@ -1,5 +1,5 @@
 import { GraphQLFloat, GraphQLString, GraphQLInt, GraphQLScalarType } from "graphql";
 
-export const Int = GraphQLInt;
-export const Float = GraphQLFloat;
-export const ID = GraphQLString;
+export const Int: GraphQLScalarType = GraphQLInt;
```
19majkel94: I don't like explicit types, I always use `noImplicitAny` and rely on type inference. Especially in this case when it's too verbose and redundant.

src/decorators/index.ts
```diff
@@ -11,4 +11,3 @@ export { GraphQLResolver } from "./GraphQLResolver";
 export { Mutation } from "./Mutation";
 export { Query } from "./Query";
 export { Root } from "./Root";
-export { Subscription } from "./Subscription";
```
19majkel94: `Subscription` decorator exist but it's not implemented yet. I would rather comment this line to prevent accidental usage of this decorator (intellisense stuff).

examples/01-simple-usage/recipe-type.ts
```diff
@@ -11,6 +11,9 @@ export class Recipe {
   @Field(type => Int)
   ratings: number[];
 
+  @Field({description: "Creation date field.", nullable: false})
```
19majkel94: Date usage example is in the `typeorm` section with `@CreateDateColumn`. Simple example should be simple, description and depreciation examples has been added already.

README.md
```diff
@@ -370,6 +370,29 @@ app.listen(4000, () => {
 
 ## Examples
 You can also check the [examples](https://github.com/19majkel94/type-graphql/tree/master/examples) folder on the repo for more example of usage: simple fields resolvers, DI Container support, etc.
+Please notice that, do tue a [ts-node bug](https://github.com/rbuckton/reflect-metadata/issues/84) an additional parameter is needed when running with ts-node:
+```bash
+ts-node --type-check examples/01-simple-usage/index.ts 
+```
+
+Query example#01-simple-usage
```
19majkel94: I would move sample queries to the specific folder (maybe as part of folder readme)

19majkel94:
@4F2E4A2E you can also set the tslint rule about enforcing spaces around `{ }` brackets 
I would like to use prettier as a commit hook but the decorators case hasn't been solved yet:
https://github.com/prettier/prettier/issues/325#issuecomment-321837796

src/scalars/aliases.ts
4F2E4A2E: it's for the readability and introduction easier phase in for new contributors

examples/01-simple-usage/recipe-type.ts
4F2E4A2E: That's confusing, since there will be two main user bases on this lib, no-typeorm interested and typeorm interested ones.
Will there be an easy and complex example for each (no-typeorm and typeorm)?

README.md
4F2E4A2E: Ok, but what do you mean with specific folder? Folder readme?

4F2E4A2E:
just created a new PR: https://github.com/19majkel94/type-graphql/pull/26

README.md
19majkel94: Yes, readme in `01-simple-usage` folder with sample queries. However I don't think it's needed, GraphiQL do it's job very well and the lib is designed for developer that know GraphQL well 😉

Issue #29, opened 2018-02-12
Bug: value.toISOString is not a function


4F2E4A2E:
I am getting following error when quering for an object with an iso date field:

```typescript
serialize(value: Date) {
  return value.toISOString();
},
```

```typescript
// date type gets lost on transpile and leads to this error
serialize(value) {
  return value.toISOString();
},
```

19majkel94:
Please provide minimal reproduction example of this bug. Looks like you are not returning `Date` instance.

4F2E4A2E:
Sorry it took so long, but here it is:
```typescript
// integration.test.date.ts
// "type-graphql": "0.4.0",
import 'reflect-metadata';
import {graphql} from 'graphql';
import * as bodyParser from 'body-parser';
import * as express from 'express';
import * as graphqlHTTP from 'express-graphql';
import {buildSchema, Field, GraphQLISODateScalar, GraphQLObjectType, GraphQLResolver, Query} from 'type-graphql';

@GraphQLObjectType()
abstract class A {
    @Field(returnType => GraphQLISODateScalar)
    a: Date;
}

@GraphQLResolver(() => TestResolver)
class TestResolver {

    @Query(returnType => A, {description: ''})
    test() {
        return {
            a: '2018-02-15'
        };
    }
}

const schema = buildSchema({resolvers: [TestResolver], dateScalarMode: 'isoDate'});

function runExpress() {
    const port = 4000;
    const app = express();
    app.use('/graphql', bodyParser.json(), graphqlHTTP({schema, graphiql: true}));
    app.listen(port, () => {
        console.log(`Graphical started, http://localhost:${port}/graphql`);
    });

    graphql(schema, `{ test{ a } }`)
        .then(f => console.log(JSON.stringify(f)));
}

runExpress();
```

**Error**:
```typescript
{"errors":[{"message":"value.toISOString is not a function","locations":[{"line":1,"column":9}],"path":["test","a"]}],"data":null}
```

19majkel94:
As I said, your code is invalid. Your not returning `Date` object but string.
When you do:
```ts
    @Query(returnType => A, {description: ''})
    test(): A {
        return {
            a: '2018-02-15'
        };
    }
```
You will get compile error that '2018-02-15' is not a `Date` object. You need to return `a: new Date('2018-02-15'),` 😉

4F2E4A2E:
[...]

PR #30, opened 2018-02-12
fix(#29): toISOString is not a function

4F2E4A2E:


4F2E4A2E:
> Wrapping date into date is not a solution

It's not elegant, but it did solve the problem and it's enforcing it the date type, which was not the case before that.

19majkel94:
> and it's enforcing it the date type

It assumes that the resolver returns date string or number timestamp. Which is incorrect, it should return `Date` instance and then everything works ok:
https://github.com/19majkel94/type-graphql/blob/tests/tests/functional/scalars.ts#L253

19majkel94:
Closing according to discussion in #29

PR #83, opened 2018-05-15
Role should be 'any' type to allow complex role definitions

alexey-pelykh:
Not every role can be defined as string, doing serialize-deserialize probably not the best idea. Inspired by `routing-controllers`


19majkel94:
Closing via 92fd5c3 🔒 
@tbrannam I've changed it to a generic, you can read the updated authorization docs for more info.

Issue #12, opened 2018-01-28
Better checking and meaningfull errors


MichalLytek:
**Please feel free to comment in this issue and tell what check is missing or what check throw not detailed errors!** 

- [x] more info about the need to provide explicit type
- [x] more info about problems with determining GraphQL type for property/argument (@esindger) 
- [x] using `@Args` with `@InputType` or without `@ArgsType`
- [x] inform about missing `reflect-metadata` polyfill
- [x] missing `@InterfaceType` on class provided to `{ implements: SomeClass }` (@gdanov)
- [x] more details in message when `GeneratingSchemaError` occured
- [ ] duplicated arg name
- [ ] description/depreciation on not supported field
- [ ] multiple usage of `@Args()`
- [ ] field resolver not matching resolver object type
- [ ] using `@Arg` with `@ArgsType` class
- [ ] duplicated query/mutation name (now it overwrites)
- [ ] overwriting interface/base class field types
- [ ] creating field resolver for method resolver
- [ ] no resolver files match glob pattern
- [ ] detect ES5 compilation (no `Function.name`)
- [ ] detect `undefined` or other invalid values provided by mistake in `type => X`
- [ ] detect wrong `target` setting and arrow function detection issue (#206, #219)
- [ ] more specific error message when root Query is missing (#218)
- [ ] using `defaultValue` option in `@Field` placed in `@ObjectType`
- [ ] abstract type is used directly in GraphQL schema
- [ ] checking implementation of all abstract interface field resolvers
- [ ] inform that decorators don't work on static methods (@karladler)
- [ ] duplicated types/enums before schema auto-check (with stacktrace to point source files)
- [ ] warn about declaring different type (or type options) in `@FieldResolver` than in `@Field()`
- [ ] ...others 😜

esindger:
>> more info about the need to provide explicit type

Type definition error text is unclear. For example, if I try to use output type in mutation, I got an error 
```
(node:10303) UnhandledPromiseRejectionWarning: Error: Cannot determine GraphQL input type for filter
```
How to figure out what exactly type was missed?

It would be nice to provide some additional info:
- resolver name
- query/mutation name
- argument name

gdanov:
**Describe the bug**
I have `CRM.Client` and `ClientImpl` with the second being annotated as

`@tq.ObjectType({ implements: CRM.Client })
export abstract class ClientImpl extends CRM.Client {`

[...]

PR #98, opened 2018-06-13
print warning if no resolver files match glob

abenhamdine:
warn if no files match the glob pattern 
would have saved me half an hour...


19majkel94:
Thanks @abenhamdine for your contribution!
Unfortunately this is a con of glob pattern loading 😞 

I think that we should throw error when no resolvers matched the pattern, as it's definitely error in the setup. So I will move this to #12, thanks anyway for pointing this issue 😉

abenhamdine:
yes, better to check globally, you're right.
it was more of a hint than a real soution.
thx for your response.

Issue #102, opened 2018-06-18
FieldResolver in abstract resolver throws an error


limenutt:
Resolver inheritance feature is great, helps to reduce the amount of boilerplate code!

Currently, if you add a `@FieldResolver` to the abstract resolver it throws an error 
_No provided object type in '`@Resolver`' decorator for class AbstractResolverClassName!'_
The easiest way to reproduce this would be to add a field resolver to the abstract resolver from the examples/resolvers-inheritance
```
@FieldResolver()
    protected async resourceName() {
      return "dummy string";
    }
```
I wonder if there is a way to add a field resolver there or the error is by design?

19majkel94:
You need to provide `of => objectTypeClass` parameter, like with a normal resolver class - check the decorator signature.

limenutt:
Oh, I see. For some reason, I thought that since the Resolver was marked with `{ isAbstract: true }`, it did not require a type, since one would be provided in the concrete implementation of this abstract resolver. Since it worked without type until I've added the field resolver to it, I decided it had something to do with field resolver. 
Thank you, after I've added the type to the abstract resolver, it works fine now.

I'd suggest updating the resolver inheritance example by supplying resolver type there as well.

19majkel94:
> Since it worked without type until I've added the field resolver to it, I decided it had something to do with field resolver.

It always worked like that - your queries and mutations don't need to be attached to specific `ObjectType` but `FieldResolver` has to. Just like it's stated in docs:
https://19majkel94.github.io/type-graphql/docs/resolvers.html#field-resolvers

19majkel94:
Fixed via 6655ac1 :tada

PR #103, opened 2018-06-18
Updated resolvers inheritance example

limenutt:
It is important to provide correct resolver type to an abstract resolver, Otherwise, field resolvers would not work.
#102


19majkel94:
Thanks @limenutt for your contribution! 😉

Unluckily, you've confused some concept of TypeGraphQL, so I can't merge it.
But soon I will update the resolvers inheritance example with field resolvers.

Issue #45, opened 2018-03-19
Custom decorators support


MichalLytek:
To improve decoupling the business logic from framework/graphql stuffs (resolvers as services), TypeGraphQL should allow to register param decorators to extract some data from context, root/source, etc. It could be for example `@CurrentUser` decorator:

```ts
const CurrentUser = createParamDecorator(({ context }) => context.user);

class SampleResolver {
  @Mutation()
  rate(@CurrentUser() user: User, @Arg("rate") rate: RateEnum) {
    this.ratingsService.addRate({ user, rate });
  }
}
```

Also it should be possible to create field/method decorators to provide some metadata or capture requests, eg. `@CacheControl`:
```ts
class SampleResolver {
  @Query()
  @CacheControl(5000)
  sampleQuery(): boolean {
    // will be called at most once per 5s
    return this.database.performCostlyOperation();
  }
}
```
This feature would be a syntax sugar over middlewares (#14) feature.

stonecold123:
Good! I've been using routing controller for a year and now I'm planning to try type-graphql because of these familiar features.  Make things a lot easier.

j:
I suppose this can also support something such as: #47

19majkel94:
As #14 has landed, field/method decorators are supported now :tada:
https://19majkel94.github.io/type-graphql/docs/middlewares.html#custom-decorators

prilutskiy:
```
@CacheControl(5000)
```
Is this something that's built in? I'd like to use it in my project

19majkel94:
With #329, now it's possible to create own custom param decorators 🚀 

@prilutskiy 
See https://github.com/19majkel94/type-graphql/blob/fdb7e57b8e08fddb0535dd601f1c5505b3a95fb3/examples/apollo-engine/cache-control.ts

Eluminati:
I have a problem with custom ClassDecorators, which blocks me now.

```javascript
// The class decorator
function baseConstructor(constParamsIndex: number = 0) {
    const original = ctor;
    // the new constructor behavior
    const f: any = function newConstructor(...args: any[]) {
        const instance = new original(...args);
        // Do magic here
        return instance;
    };
    // copy prototype so intanceof operator still works
    f.prototype = original.prototype;
    // Copy static members
    Object.keys(original).forEach((name: string) => { f[name] = (<any>original)[name]; });
    // return new constructor (will override original)
    return f;
}

// The ObjectModel
@baseConstructor() @ObjectType()
[...]

PR #113, opened 2018-07-11
Pass resolver's data/request context to DI container

scalder27:
This will allow using context dependant injections. This is extremely useful for tracking logs by individual requests.

The library is really cool. One thing that is preventing me from using it - inability to pass request context to DI Container. 

Example
```ts
import { useContainer, buildSchema } from "type-graphql";
import { createContainer, asClass } from "awilix";

import { SampleResolver } from "./resolvers";

const container = createContainer().register({
  SampleResolver: asClass(SampleResolver),
});

// register the 3rd party IOC container
useContainer({
  get(someClass: any, resolverData: ResolverData<{requestId: string}>) {
    const scopedContainer = container.createScope().register({
      requestId: asValue(resolverData.context.requestId),
    });

    return scopedContainer.resolve(someClass.name);
  }
});

// build the schema as always
  resolvers: [SampleResolver],
});
```


19majkel94:
I have thought about this design earlier but I resign from implementing it because I haven't found a way to enforce unique context/token to scope the container to individual graphql request on a schema level. Of course you can provide function to `graphql-yoga` or `express-graphql` and have it but someone may forget about it and have issues in debugging the resolvers/services that shouldn't be singletons. It's complicate the design and makes too much magic.

Secondly, I think that the container per request idea is against the spirit of single-threded, event-loop based Node.js ecosystem. By creating all services/controllers/resolvers per request we goes straight to the "request -> process spawn -> response -> process died". It will will have huge impact on memory usage and total active connection (remember about graphql subscriptions that keeps alive, they're not closed after a few seconds like http request).

We already has this discussion in rest framework:
https://github.com/typestack/routing-controllers/issues/174
My conclusion is that the services should be stateless, not scoped to user/request. All the "dynamic" data should be normally passed by params explicitly.

> This is extremely useful for tracking logs by individual requests.

I see what's your point. You want some magic to implicitly under the hood pass the data to service, instead of passing it by hand:
```ts
@Resolver()
export class UserResolver {
 constructor(private logger: Logger) {}

 @Query()
 getAll(@Ctx() ctx: Context) {
   this.logger.log("message", ctx);
 }
}
```
I think the solution for this problem might be #45 but it will work only with `Resolver` classes.
```ts
@Resolver()
export class UserResolver {
 @Query()
 getAll(@InjectLogger() logger: Logger) {
   this.logger.log("message");
 }
}
```

I will have to think how to implement this wisely, maybe using `node-continuation-local-storage`. For now I will have reimplement this PR (or push commits to your branch) without lodash and using TypeDI in tests. And if I don't find a better way, I will make this feature official by providing examples and proper docs.

scalder27:
Hi @19majkel94!

Sorry I didn't write answer earlier. My bad.

About using context. Of course, there is no way to enforce context on a schema level. It's a server's responsibility to inject it. But as far as I know, there are plenty of tools to help with it. For example, [recommended in graphql-js docs](express-graphql) `express-graphql` is passing the whole request to the resolver's context by [default](https://github.com/graphql/express-graphql#options):
```markdown
context: A value to pass as the context to the graphql() function from GraphQL.js/src/execute.js. If context is not provided, the request object is passed as the context.
```

Maybe someone will forget, maybe for someone, it's unnecessary complex. But it's okay. Developers building complex things using complex (complicated sometimes) tools. Someone may forget anything at any time. It's impossible to avoid completely. But the good thing about context in resolvers - you have to specifically set it up. It's like the settings for an advanced user that knows what he wants to get. For the rest - reasonable default behavior, singletons, and no problems at all.

About single-threaded spirit. Sorry, but why it's even an argument? A scoped container has nothing to do with threads. And it's doing no more harm to NodeJS way then DI in general. Sure it's more expensive for the performance and memory, and it's okay to warn about it. But who could guess user's budget for memory, space, amount of clusters, how much nodes he could spin? It's quite possible that ability to have better monitoring for all servers means much more for him. At least for me, knowing impact on memory and getting all the benefits of "requestId" in logs, it's a tradeoff worth to make. And, actually, passing context to container resolver has no impact on performance in the context of `type-graphql`.

If this discussion has happened in the context of 'typedi' library, I'd say, sure, you right, you think it's a wrong way to make containers and you don't have to force yourself to create a scoped container. But, we're talking about `type-graphql`. The library, that promise support for (any) containers, and saying nothing about incompatibility with certain types of them. It's just wrong. Any user using "wrong" type container will just waste time trying to fit the library in a project.

The solution with passing log's related context with "requestId" is something. If a server is anything a bit more complex than a wrapper over DB, there will be a lot of methods. I was doing it, it's awful. A lot of noise, a lot of tedious and useless work. I'd prefer to stay away from the library, then coming back to such a solution. 

How about to add an explicit option to pass context to containers. It will be turned off by default. And you'll warn in the documentation to use it only if it's REALLY necessary. I'll remove all unnecessary dependencies like lodash and uuid, but I'd prefer to have a test with a scoped container. And TypeDI doesn't suit well for this. Of course, if you don't mind an option thing and passing context to containers.

Again, thank you for the library! DI or not DI, it was a pleasant acquaintance!

19majkel94:
I understand your point. It's easy to pass the context to container's `get` function and it has no performance impact. But my doubts are about implementing any weird stuffs that results in complete mess in user's codebase. If there's too many options to achieve some result, it's easy to hurt yourself and blame this stupid framework.

I just want to balance between universality/elasticity and being opinionated. I try to analyze all use cases that user's report and find the best solution for the problem. The solution that fit to the idea and is clean and simple.

> It's like the settings for an advanced user that knows what he wants to get. For the rest - reasonable default behavior, singletons, and no problems at all.

I get it but from my experience, a lot of the people don't read the docs. They will just copy the code from examples and get it troubles by instancing the whole world for each request.

> About single-threaded spirit. Sorry, but why it's even an argument? A scoped container has nothing to do with threads. And it's doing no more harm to NodeJS way then DI in general. Sure it's more expensive for the performance and memory, and it's okay to warn about it. 

Node.js is single threaded so there's no thread overhead for each request like in PHP. You don't have to instantiate the whole app for each request like php frameworks do by Apache, so you can handle hundreds of thousands parallel connections (websockets). If you have scoped containers that creates all services, resolvers and repositories for each request (or websocket connection), you loose the main benefit of Node.js - great resources utilization. You will pollute the memory with plenty of instances just for implicitly passing some context data.

> But who could guess user's budget for memory, space, amount of clusters, how much nodes he could spin?

More and more developers recently are switching from Node.js to Go for large scale microservices that works on clusters. So I definitely have to profile this feature to get know how much impact it has for http and subscriptions and warn about it in docs.

> I'll remove all unnecessary dependencies like lodash and uuid, but I'd prefer to have a test with a scoped container. And TypeDI doesn't suit well for this. 

`TypeDI` has `Container.of` feature for scoped containers and `@Service({ global: true })` for pure services (singletons), so it's not a problem. I will try to play with it this week :wink:

scalder27:
Hi @19majkel94,

Didn't know `TypeDI` supports scoped containers! Worth to check it =)

It's an interesting thing that all your reasoning centered in preventing dumb users to hurt themselves in the areas outside the library's responsibility. Usually, reasonable default settings + rich customization suits for users better. Take for example, `react-create-app` that has the easiest default settings and, simultaneously, has the power to customize set up however suits for a user's need. It's not trying to cripple a fellow developer.

Okay, please, have a look, have a thought, check the performance ;)

I'll try to replace `awilix` with `TypeDI` and will get rid of unnecessary dependencies.

19majkel94:
> It's an interesting thing that all your reasoning centered in preventing dumb users to hurt themselves in the areas outside the library's responsibility. Usually, reasonable default settings + rich customization suits for users better. Take for example, react-create-app that has the easiest default settings and, simultaneously, has the power to customize set up however suits for a user's need. It's not trying to cripple a fellow developer.

Ideally, I would like to have a slim basic core and all integration and not common features would be a plugin. There would be a `@typegraphql/common` package that gathered common plugins in one product + `typegraphql` CLI for generating the project like `create-react-app` 😉

scalder27:
@19majkel94, sounds good!

19majkel94:
@scalder27 Please take a look at the example of TypeDI integration with scoped containers:
https://github.com/19majkel94/type-graphql/tree/79321593b9719646dcacb30f5fa72a72f4e6162b/examples/using-scoped-container

19majkel94:
Right, thanks for pointing that!
I forgot about it because again, it's hard to hook into the last field resolver.

Your solution with middleware (graphql-middleware or TypeGraphQL middleware) has no memory leaks but it creates a fresh scoped container for each resolver, even field resolver.

I have to figure out how to hook into this lifecycle. The only solution might be creating own express/koa middleware, like vesper makes a wrapper:
https://github.com/vesper-framework/vesper/blob/68ba63931a6d3f370463e51321fe2902dd4fcf4e/src/index.ts#L99

scalder27:
@19majkel94 hi!

Yeah, sure you right about graphql middlewares.
Based on `graphql-yoga` docs it should be something like:

```
const server = new GraphQLServer({
    schema,
    // we need to provide unique context with `requestId` for each request
    context: (): Context => ({
      requestId: Math.floor(Math.random() * 1000000), // uuid-like
    })
});

// keep in mind that using scoped container could be dangerous, sometimes you have to take care of freeing memory.
server.express.use(disposeScopedContainerMiddleware);
```

19majkel94:
I'm not sure whether graphql-yoga support `after` express middlewares.

But thankfully the `formatResponse` option works great!
```ts
formatResponse: (response: any, { context }: any) => {
  console.log((Container as any).instances.map((it: any) => it.id));
  Container.reset(context.requestId);
  return response;
},
```

I've tested it with `sleep(10000)` in resolver and 3 concurrent requests. The `console.log` printed:
```
[ 866135, 427927, 73156 ]
[ 427927, 73156 ]
[ 73156 ]
```

Anyway, it looks complicated end error prone.
I think I would have to create my own `GraphQLServer` implementation in `@typegraphql/http` or common, where I would expose the before/after request hooks. It allows me to control the query execution so I would be able to support document-level middlewares that would be called once per request, not once per resolver (multiple times per query). So it would be then soooo easy to create a middleware for creating and destroying scoped container.

scalder27:
`formatResponse` hacky as hell ;)
but I don't know a good 100%-bulletproof way for after-request middleware in express

scalder27:
Ah, wait, something like this should work:

```typescript
const createContextMiddleware = (req, resp, next) => {
  req.context = {
    requestId: Math.floor(Math.random() * 1000000), // uuid-like
  };
  next();
};

const disposeScopedContainerMiddleware = (req, resp, next) => {
  const context = req.context;
  req.once("end", () => {
    if (Container.has(context)) {
      Container.reset(context)
    }
  });
  
  next();
};

const server = new GraphQLServer({
  schema,
  context: (req): Context => req.context
});

server.express.use(createContextMiddleware);
server.express.use(disposeScopedContainerMiddleware);
```

19majkel94:
@scalder27 
Thanks, `res.once("finish"` works! 😃 But again we need an express middleware for injecting context into `req` object to retrieve it in `disposeScopedContainerMiddleware`.

So I will leave the `formatResponse` workaround for now, until `documentMiddlewares` arrive:
https://github.com/prismagraphql/graphql-yoga/issues/314

19majkel94:
@scalder27 
I just figure out that we don't need that changes and pull request at all.
Scoped container means that there is a separate instance of container for each request. And to make it work per request, we have `createContextMiddleware`. So we can easily create there an instance of container base on `requestId` and attach it to `context` object.
```ts
const createContextMiddleware = (req, res, next) => {
  const requestId = Math.floor(Math.random() * 1000000); // uuid-like
  req.context = {
    requestId,
    container: Container.of(requestId),
  };
  next();
};
```

And to avoid the need of creating a wrapper to use in `useContainer`:
```ts
// create a wrapper for TypeDI `Container.of` scoped container feature
export const ScopedContainer = {
  get(someClass: any, { context }: ResolverData<Context>) {
    // use context to get scoped container
    const container = context.container;

    // then get from container like normal
    return container.get<any>(someClass);
  },
};
```
I would rather expand its definition to pass a function to extract the container from resolver data:
```ts
useContainer(({ context }) => context.container);
```

Sorry for the delays in working on this feature that is a must-have for you but I need to ensure that the API is the best I can achieve to prevent future refactoring and breaking changes 😉

scalder27:
@19majkel94,

It's a really good idea! More flexible, more predictable!

19majkel94:
Closed via #121 and published in v0.13.0 🎉

scalder27:
@19majkel94 thank you very much!

Issue #125, opened 2018-08-13
Support entity inheritance


artman:
In order to maintain a large number of entities it would be beneficial to be able to use inheritance to not have to redefine entity fields.

The following demonstrates this:

```
class BaseEntity {
  @Field({ description: "The identifier of the entity." })
  public id: string;

  @Field({ description: "The time at which the entity was created." })
  public createdAt: Date;

  @Field({ description: "The time at which the entity was updated." })
  public updatedAt: Date;
}

@ObjectType({ description: "A user" })
export User extends Model {
  @Field({ description: "The name of the user." })
  public name: string = "";
}
```

artman:
My bad. Decorating the base class with ObjectType will make this work.

PR #132, opened 2018-08-24
Clarify object type inheritance docs

davewasmer:
See #125 for an example of the confusion

### DIFF
--- a/docs/interfaces-and-inheritance.md
+++ b/docs/interfaces-and-inheritance.md
@@ -70,13 +70,20 @@ class GetTodosArgs extends PaginationArgs {
 
 This technique also works with input type classes, as well as with object type classes:
 ```typescript
-// `Person` is the object type class we've created earlier in this docs
+@ObjectType()
+class Person {
+  @Field()
+  age: number;
+}
+
 @ObjectType()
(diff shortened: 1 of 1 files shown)


docs/interfaces-and-inheritance.md
```diff
@@ -70,14 +70,21 @@ class GetTodosArgs extends PaginationArgs {
...
+}
+
 @ObjectType()
 class Student extends Person {
   @Field()
   universityName: string;
 }
 ```
 
+Note that both the subclass and the parent class must be decorated with the `@ObjectType()` decorator for the parent class fields to be picked up on the subclass.
```
19majkel94: Note that both the subclass and the parent class must be decorated with the the same class decorator (e.g. `@InputType()` -> `@InputType()`) for the parent class fields to be picked up on the subclass. This means that you can't create e.g. input type class that inherits from object type class.

davewasmer: Updated to reflect this as well

19majkel94:
The docs are live! 🎉 
@davewasmer Thank you very much for your contribution! ❤️

Issue #156, opened 2018-09-23
[FR] Default naming strategy 


stevefan1999-personal:
**Is your feature request related to a problem? Please describe.**
I'm currently frustrated at doing this repetitively:
```ts
...
  @Field(() => Boolean, { name: 'have_children' })
  get haveChildren (): Promise<boolean> {
    return (async () => ((await this.children) || []).length > 0)()
  }

  @Field(() => Boolean, { name: 'is_children' })
  get isChildren (): Promise<boolean> {
    return (async () => (await this.parent) !== undefined)()
  }

  @Field(() => Boolean, { name: 'is_root' })
  get isRoot (): Promise<boolean> {
    return (async () => (await this.parent) === undefined)()
  }
...
```
And this:
```ts
...
  @Query(() => Submission, {
    nullable: true,
    name: 'submission_feed'
  })
  async feed (@Arg('id') id: number): Promise<Submission> {
    return this.submissions.findOneOrFail({ id }).fail('No such submission')
  }

  @Authorized() @Mutation(() => Submission, { name: 'submission_submit' })
  async submit (
    @Args() { url, description }: SubmitArgs,
    @Ctx('user') submitter: User
  ): Promise<Submission> {
    // secret code...
  }
...
```

So we're basically using camelCase in server code but we also decided to use snake_case for the name convention of our GraphQL schema so that we could get a better separation of namespaces (because GraphQL doesn't support nested queries yet, we had to resort back to oldschool C-style OOP namings)

**Describe the solution you'd like**
Let there be a default naming strategy transformer in the method `buildSchema` that is undefined by default. It will be a/an (a)synchronous function with arguments `(className: string, name: string, type: 'field' | 'resolver' | 'query') => string | Promise<string>`, then after the basic schema skeleton is built, we do a DFS on that, then we will convert the name into various cases, for example, by using [blakeembrey/change-case](https://github.com/blakeembrey/change-case).
If a `name` property is already denoted on the decorator metadata, name transformation will be ignored.

**Alternative solution**
It seems like we could also do this externally ourselves by traversing the schema ourselves, although it will be harder to determine contexts and types and such due to loss of metadata.

19majkel94:
Thanks for sharing your idea! 😃 

I like it, it's implementable, so I've added it to the features list 😉

> because GraphQL doesn't support nested queries yet

What do you mean? It might be not semantically correct but you can implement queries as a field resolvers:
```graphql
query NamespaceExamples {
  submission {
[...]

PR #157, opened 2018-09-23
Implements name transformers #156

stevefan1999-personal:
⚠Not available for merging yet⚠

Implements #156

Some problems to face now:
- [ ] Not yet tested ~(but should work out of the box)~
- [ ] Eliminate global state hazard
- [ ] (Optional) use a better strategy (currently: transforms names before schema generation)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,9 +19,9 @@
       "integrity": "sha512-q4TfI+jJISul6vVpZJktzH4tupwRiVk6KXRhB8PHqJ7erl966I6ePDXl9mAbE8jMM7YswhnnB0j1SYP7LBVyhg==",
       "dev": true,
       "requires": {
-        "chalk": "2.3.0",
-        "esutils": "2.0.2",
-        "js-tokens": "3.0.2"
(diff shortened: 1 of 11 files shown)


19majkel94:
Sorry but #156 in term of custom naming strategy function with metadata access is a bit complicated and I should do it by myself after MetadataStorage redesign (I've described it  in https://github.com/19majkel94/type-graphql/issues/133#issuecomment-415982775).

If you want to help me, please add some examples of usage of the naming strategy function API in #156 to better show me what you expect from this. I've seen you've edited it in the meantime so please focus on that part.

I am closing this PR as it's a not functioning draft. Please open a new one when it will be finished and you want to show me what you expect 😉

stevefan1999-personal:
But as you can see, it is largely working🙃Just need some more refinement to make it to the next metadata design.

For this matter, I dreamt on how about associating each metadata into the class function itself? And when we build schemas we collect from those classes. The downside of this is that we had to exhaustively select all classes to collect data from, and so glob support is possibly gone.

Issue #124, opened 2018-08-11
Decorator allowing field metadata annotations


chrisdevereux:
It is generally useful to be able to annotate fields with custom metadata for use by middleware and validators. See: #77 and #36 (my personal use case)

Since there currently isn't support for directives in GraphQL-js, an alternative might be for TypeGraphQL to add a new decorator that annotates the fiend config object passed into the GraphQLObjectType constructor with additional user-defined metadata fields.

I'd imagine the API looking something like this:

```typescript
@ObjectType()
class MyObjectType {
  @Metadata({ isUseful: false })
  @Field()
  myAnnotatedField() {
    return "12"
  }
}
```

This relatively simple change would make it possible to write middleware libraries wrapping tools such as graphql-query-complexity.

> **Owner note:**
> Read more about how to implement here: https://github.com/19majkel94/type-graphql/issues/124#issuecomment-426935908

19majkel94:
To access a data from decorator in middleware, you can create custom decorators that uses middlewares under the hood:
```ts
function Metadata(args: { isUseful: boolean}) {
  return UseMiddleware(({}, next) => {
	if (!args.isUseful) {
	  return null; // or other logic	
    }
    return next();
  });
}
```

If you need to access graphql-related metadata, you might be interested in #123.

If you think about #36 and the need to emit some data into `new ObjectType` constructor, it's a separate thing. I think i it may be implemented only with the plugin API, as I need to call it on schema creation, it can't be done by a middleware.

chrisdevereux:
Yeah, I'm thinking this would be applied at schema creation, then a global middleware would have access to the annotated schema and would be able to work it's magic.

19majkel94:
So you also need `graphql-js` related metadata like `GraphQLInputObjectTypeConfig` data? For #36 you only need to inject the data into the config on build types time.

Sorry but I don't understand the _"middleware would have access to the annotated schema"_ part 😕

19majkel94:
> How would you feel about integrating this feature in the core since there isn't a plugin API yet?

For #36 I would have to add a field to decorator config, store it in MedatataStorage and manually place in ObjectType config object on schema generation.
But I don't like this tight coupling, I like the idea of microservices and I would prefer to go with multi packages monorepo. To do this, I need to develop a plugin API, so `@typegraphql/query-complexity` would be installed by user, registered in `buildSchema` `plugins` array and extend the `@Field` decorator with `complexity` settings.
[...]

PR #163, opened 2018-10-08
Added basic implementation for #124

wesselvdv:
Based on the issue #124, I added some very basic tests let me know if more are desired.

### DIFF
--- a/src/decorators/Metadata.ts
+++ b/src/decorators/Metadata.ts
@@ -1,0 +1,27 @@
+import { getMetadataStorage } from "../metadata/getMetadataStorage";
+import { MethodAndPropDecorator } from "./types";
+
+export interface Metadata {
+  [index: string]: any;
+}
+
+export function Metadata(data: Metadata): ClassDecorator | MethodAndPropDecorator | any {
+  return (
+    targetOrPrototype: Function | Object,
+    propertyKey?: string | symbol,
+    descriptor?: any,
+  ): void => {
+    if (typeof targetOrPrototype === "function") {
+      getMetadataStorage().collectAdditionalObjectTypeMetadata({
+        target: targetOrPrototype,
+        data,
+      });
+    } else if (propertyKey !== undefined) {
+      getMetadataStorage().collectAdditionalFieldMetadata({
+        name: propertyKey,
+        target: targetOrPrototype.constructor,
+        data,
+      });
+    }
+  };
+}
--- a/src/metadata/definitions/class-metadata.ts
+++ b/src/metadata/definitions/class-metadata.ts
@@ -6,4 +6,5 @@ export interface ClassMetadata {
   fields?: FieldMetadata[];
   description?: string;
   interfaceClasses?: Function[];
+  metadata?: any;
 }
--- a/src/metadata/definitions/field-metadata.ts
+++ b/src/metadata/definitions/field-metadata.ts
@@ -15,4 +15,5 @@ export interface FieldMetadata {
   params?: ParamMetadata[];
   roles?: any[];
   middlewares?: Array<Middleware<any>>;
+  metadata?: any;
 }
--- a/src/metadata/definitions/index.ts
+++ b/src/metadata/definitions/index.ts
@@ -6,3 +6,4 @@ export * from "./middleware-metadata";
 export * from "./param-metadata";
 export * from "./resolver-metadata";
 export * from "./union-metadata";
+export * from "./metadata";
--- a/src/metadata/definitions/metadata.ts
+++ b/src/metadata/definitions/metadata.ts
@@ -1,0 +1,5 @@
+export interface Metadata {
+  data: any;
+  target: Function;
+  name?: string | symbol;
+}
--- a/src/metadata/metadata-storage.ts
+++ b/src/metadata/metadata-storage.ts
@@ -1,25 +1,26 @@
+import { NoExplicitTypeError } from "../errors";
+import { ClassType } from "../interfaces";
 import {
-  ResolverMetadata,
-  ClassMetadata,
-  FieldMetadata,
-  ParamMetadata,
-  FieldResolverMetadata,
   AuthorizedMetadata,
   BaseResolverMetadata,
+  ClassMetadata,
   EnumMetadata,
+  FieldMetadata,
+  FieldResolverMetadata,
+  Metadata,
+  MiddlewareMetadata,
+  ParamMetadata,
+  ResolverClassMetadata,
+  ResolverMetadata,
+  SubscriptionResolverMetadata,
   UnionMetadata,
   UnionMetadataWithSymbol,
-  ResolverClassMetadata,
-  SubscriptionResolverMetadata,
-  MiddlewareMetadata,
 } from "./definitions";
-import { ClassType } from "../interfaces";
-import { NoExplicitTypeError } from "../errors";
 import {
-  mapSuperResolverHandlers,
+  ensureReflectMetadataExists,
   mapMiddlewareMetadataToArray,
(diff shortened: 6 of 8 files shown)


wesselvdv:
Why is the ci failing? All I can see is some comment about a time out, I can't reproduce this locally.

19majkel94:
> Why is the ci failing? 

I thought that's new jest setup issue but looks like it's travis fault. I've fixed it on the branch https://github.com/19majkel94/type-graphql/pull/167/commits/78990abeee6ec52bd693386c95b6f2b11533c679, wait for the merge and then rebase your PR.

wesselvdv:
I’m not sure I follow? I need to store the custom object somewhere where the chance it conflicts with other parameters is the lowest. (e.g. to prevent accidental key/paramter collisions.) I am however using object destructuring to add the entire metadata object to the type config on creation. Which is what I believe you also suggested no?

wesselvdv:
That could definitely work, but how to get it on the type config then? For example for joinmonster we really need specific parameters to exists in the typeconfig, not in the metadata storage.

19majkel94:
Runtime example
```ts
// type metadata definition
@ObjectType
class MyType {
  @Metadata({
    complexity: 5,
  })
  @Metadata({
    isTopSecret: true,
  })
  @Field()
  field: FieldType;
}

// metadata storage
metadata: [
  {
    complexity: 5,
  },
  {
    isTopSecret: true,
  },
],

// GraphQLObjectType
fields: () => {
  // ...
  complexity: 5,
  isTopSecret: true,
},
```

I can't explain that simpler 😆

wesselvdv:
I believe we’re talking about different things, in your approach could I theoretically get that custom metadata from the schema definition contained in the resolveinfo of a resolver? (e.g. using the @Info decorator?)

19majkel94:
But you are actually doing the same 😆 	
https://github.com/19majkel94/type-graphql/blob/85b3f47472f67ed27067f459ff00bae2f27638b2/src/schema/schema-generator.ts#L243-L245

It just should be `...field.metadata.reduce((obj, metadata) => ({...obj, ...metadata}), {})` and for every type, not only object type. And testing metadata storage without checking the actual emitting in schema makes no sense.

wesselvdv:
I can agree with the testing part. Any pointers you can give me there, other than loose statements and bits of code? 

I took the liberty to quote your comment from #124:

> @wesselvdv
> Not yet but I think it's easy to implement even by beginners 😉
> 
> API usage - `@Metadata`:
> 
> ```ts
> @ObjectType()
> class MyObjectType {
>   @Metadata({ isUseful: false })
>   @Field()
>   myAnnotatedField() {
>     return "12"
>   }
> }
> ```
> 
> Steps to do:
> 
> * create Metadata decorator that accepts object
> * modify field/query/mutation/subscription/input metadata interfaces in storage
> * create a method in metadata storage that will be called by `Metadata` to place the metadata object in storage
> * use that metadata object during schema construction in all object types, input, handlers, etc.:
> 
> ```ts
> fieldsMap[field.schemaName] = {
>   // ...
>   deprecationReason: field.deprecationReason,
>   ...(field.metadata || {}),
> };
> ```
> 
> * of course test this feature well and create documentation for it 😉

I believe that what I did is what you tried to describe there? Difference is that I only created it for a field and an object type.

From what I understand from you now, you would rather collect the metadata in an array, and then consolidate them in the corresponding type on schema creation? (eg. the reduce action with double destructuring)

> The purpose of @Metadata decorator is to emit fields in GraphQL type config, just like complexity 
> does, not to place metadata key.
This comment I really don't understand, I need to store the metadata somewhere? I can't just slap it on the corresponding type its metadata storage, I believe that would get messy and it'd be easy to start messing with the internals of type-graphql unwittingly. Besides you yourself keep referring to a metadata property on a given type its metadata storage.

19majkel94:
I see that we're not able to make up about this 😄 
I have no time to deliberate on this feature any longer, looks like you have to wait until I implement it in my own way, sorry 😉

wesselvdv:
Alright, your call...

19majkel94:
Sorry, but I've wrote it a few times and I've even [showed an example](https://github.com/19majkel94/type-graphql/pull/163#issuecomment-430008214) that it should  allow for multiple decorators, so you just have to store an array in metadata storage and flatten it in schema generator. Quoting my old posts about old designs what improve this PR magically 😜 Also while going stick to the #124 guideline you've forgot about "in all object types, input, handlers, etc.".

And by:
> is to emit fields in GraphQL type config, just like complexity does, not to place metadata key

I was referring to the tests when you only check the metadata storage, not the e2e emitting properties in object type instance 😉 

Is it just a little bit more clear now? 
And sorry but I have no time to guideline every contributor and make a checklists of all steps for a simple PR. We can iterate on this PR if you know what and how to do, otherwise it's waste of time that I can use just by implementing it.

wesselvdv:
> Sorry, but I've wrote it a few times and I've even [showed an example](https://github.com/19majkel94/type-graphql/pull/163#issuecomment-430008214) that it should allow for multiple decorators, so you just have to store an array in metadata storage and flatten it in schema generator. 

Explaining something badly in a couple of comments doesn't magically make it a good explanation. You leave out a lot of stuff that happens in between your small examples, and this stuff might not have to be mentioned to you or any long time contributor, but for someone like me who hasn't memorised your source code it can be hard to follow.

Your "example" is a snippet where metadata is suddenly shown in the fields object instead of being properties of a field. 

> ```typescript
> fields: () => {
>   // ...
>   complexity: 5,
>   isTopSecret: true
> }
> ```

vs

> ```typescript
> fields: () => {
>   // ...
>   field1: {
>       complexity: 5,
>       isTopSecret: true,
>   }
> }
> ```

I assume that the latter is what you expect? That's what I would expect when adding metadata to a field type.

> Quoting my old posts about old designs won't improve this PR magically 😜 Also while going stick to the #124 guideline you've forgot about "in all object types, input, handlers, etc.".

Quoting your old post was just to show you why I chose the approach I did, and I already said that I didn't implement it for other types. The reason for that is twofold, one for the fact that I didn't need it for joinmonster, and the other for reducing the amount of stuff I would have to refactor should it have to work differently.

> And by:
> 
> > is to emit fields in GraphQL type config, just like complexity does, not to place metadata key
> 
> I was referring to the tests when you only check the metadata storage, not the e2e emitting properties in object type instance 😉

Alright, that makes sense.

> 
> Is it just a little bit more clear now?
> And sorry but I have no time to guideline every contributor and make a checklists of all steps for a simple PR. We can iterate on this PR if you know what and how to do, otherwise it's waste of time that I can use just by implementing it.

It's definitely more clear now, but what might a "simple" PR for you or someone who's been working on this project for a while, can be a different matter for someone who isn't. If you're not willing or capable to explain yourself more thoroughly when asked, then you shouldn't let the community assist you. You can't expect people to think exactly the same way you do, we're all trying our best and we're not here to annoy you. So for the sake of helping contributors a set of guidelines might be super useful. So instead of having this discussion with me or any other type-graphql noob, you'd only have to refer to your guidelines. Makes everyone's lives easier. 😉

I was just trying to (partially) implement it the way you dictated in #124, and while I am definitely willing to accept criticism, you should know that the way you communicate has a hint of a condescending tone. Comments such as "I think it's easy to implement even by beginners 😉", "Quoting my old posts about old designs won't improve this PR magically 😜" or "I can't explain that simpler 😆" won't motivate people to assist you. I am not sure if this is what you intend, but I don't think it's the best way to get people to help you.

wesselvdv:
Alright, awesome! I'll reopen this one when I have some time to rework it to what you want.

Issue #142, opened 2018-09-02
Pagination model


19majkel94:
I think it might be nice to have a pagination model that will allow to create uniwersal resolvers, types definitions, etc. for collections.

For now it's just an idea, I need to investigate how much logic I can hide without sacrificing flexibility and how the API should look. You can also paste some examples how you solved that problem with apollo/graphql-tools or with TypeGraphQL.

I think there should be two options available:

### Simple pagination
Like with standard rest apis - skip/offset, takie/limit, count, items, etc.

### Relay connection
Using cursor model, which is better for highly dynamic collections:
https://facebook.github.io/relay/graphql/connections.htm

erencay:
@19majkel94 Sorry if this is mostly irrelevant but recently I started using dedicated query and mutation resolvers which extends abstract resolver classes with an abstract `resolve` method. Under the hood they can hide a lot of boilerplate and provides a clean implementation. Each `class` represents one and only one operation with a naming strategy based on the class name. I have yet to implement  this solution for pagination but it looks like this approach works the best for me, so far. 

The thing is, this is something resembles may be what a "conventional framework" would do? I don't know if this kind of approach is beyond `type-graphql`'s intention. Also `typegraphql` is successfully using the decorator pattern which you can simply plug-in/out features. Inheritance might not have it's place in it to solve such problems in the first place.

Just wanted to point out how I'm planning to solve the problem.


```
@InputType()
class LoginInput {
    @Field()
    public username: string;

    @Field()
    public password: string;
}

@MutationResolver(LoginInput, type => Login)
class Login extends AbstractMutationResolver<LoginInput> {
    public async resolve() {
        if (this.ctx.currentUser) {
            throw new Error("You are already logged in.")
        }

        return this.input.username == this.input.password;
    }
}
```

19majkel94:
I'm glad that you found a way to reduce a boilerplate for your case 😃 
I would love to take a look at it when you finish it.

But I think that such constructs that you've presented is way too much complicated design to be the official API. Of course, it can reduce boilerplate but also introduces a huge amount of magic which is not intuitive.

TypeGraphQL design goal is to be simple yet powerful. The decorators to schema idea is understandable, features like authorization or validation are powerful yet simple. So do the pagination API has to be simple and easy to use, like the one in https://github.com/indigotech/graphql-schema-decorator.

tonyxiao:
[...]

PR #171, opened 2018-10-17
[WIP] Connections support

orta:
Hi there! I'm not yet at implementation stage for this, but I do want to try and spec out what the client code would look like (as that's the easiest to discuss and the biggest variable for me)

To do connections right we'll need to be able to:

- Define a ConnectionType (and allow subclassing with new root fields)
- Define a connection query (to ensure the 4 inputs that have to be there are in, but also to allow expanding with your own args)  
- Define a custom edge in your connection (to allow adding metadata to the connection)
- Define the node for your connection (That's kinda the entire point of a connection)

So me + @l2succes tried speccing out two versions of how you could set it up. One with all the defaults, the simplest case. The other with every possible change. We have a mix of both in our existing codebases, so most usage falls somewhere in-between.

What do you think?

### DIFF
--- a/examples/connections/advanced/recipe-advanced.type.ts
+++ b/examples/connections/advanced/recipe-advanced.type.ts
@@ -1,0 +1,103 @@
+import { Field, ObjectType, Int, ArgsType } from "../../../src";
+import { ConnectionType } from "../../../src/decorators/Connection";
+
+// The default classes that a user would pull out of this library
+
+class Edge<NodeType = {}> {
+  @Field()
+  node: NodeType;
+  @Field()
+  cursor: string;
+}
+
+class Connection<NodeType = {}, EdgeType = Edge<{}>> {
+  @Field()
+  edges: [EdgeType];
+
+  @Field() // TODO, fill out this object
+  pageInfo: any;
+}
+
+@ArgsType()
+class ConnectionArgs {
+  @Field(type => Int, { nullable: true })
+  first?: number;
+
+  @Field(type => Int, { nullable: true })
+  last?: number;
+
+  @Field({ nullable: true })
+  after?: string;
+
+  @Field({ nullable: true })
+  before?: string;
+}
+
+// Client code
+
+// A typical User type that has a connection for their recipes.
+@ObjectType()
+export class User {
+  @Field()
+  name: string;
+
+  @Field(type => RecipeConnection)
+  recipes: RecipeConnection;
+}
+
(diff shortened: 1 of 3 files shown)


19majkel94:
Hi! 👋 
This looks good, I'm glad that you like the `Recipe` example 😄 

The problem is with generic types that uses reflection and decorators. You can't just provide `T` type to the decorator, and the syntax like:
```ts
class Edge<NodeType = {}> {
  @Field()
  node: NodeType;
  @Field()
  cursor: string;
 }
```
just won't work - reflection says that type of `node` property is `Object`. So I think that the `@ConnectionType` decorator is only to provide the `node` and `edgeType` info, right?

To achieve generic types with decorators, you have to create higher order class - a function that will take the runtime class as parameters and return new class, just like with resolvers classes:
https://19majkel94.github.io/type-graphql/docs/interfaces-and-inheritance.html#resolvers-inheritance

It's not a big deal but we have to add the `isAbstract` option to prevent emitting it in schema (or maybe it just works? I don't know honestly 😄). And then you just call it in a slightly different way: - `extends Edge(User)` instead of `extends Edge<User>`.

I think that if user need some additional metadata in connection or edge, he should create the classes by itself by extending the base built-in classes, just like in your advanced example:
```ts
export function <TNodeType>Edge(NodeType: TNodeType) {
  abstract class EdgeType {
    @Field(type => NodeType)
    node: TNodeType;
    @Field()
    cursor: string;
  }
  return EdgeType;
}

@ObjectType()
export class RecipeEdge extends Edge(Recipe) {
  @Field()
  personalNotes: string;
}
```
And then create his own `RecipeConnection` and `RecipeConnectionArgs` classes to handle the pagination by itself.

But for the simple example, I would make it as less verbose as it can be - just providing the information about using the pagination like `@ConnectionArgs` and the rest (node type) can be detected from existing decorators, without the need of manual subclassing. There should be also some built-in helpers and types for creating the connection response.

I think that it's a bit too complex feature to let newcomers make a PR as I'm not 100% sure about the API and can't provide precise tips about implementation without proper research and experimenting by my own 😕

tonyxiao:
Related issue : https://github.com/19majkel94/type-graphql/issues/142

orta:
Thanks for the feedback @19majkel94 - that's super reasonable, I think I agree that maybe this one might be a bit tricky for a first timer, gonna close it and keep an eye on #142 👍

PR #148, opened 2018-09-09
chore(deps) move graphql into peerDependecies

capaj:
fixes #144

Issue #144, opened 2018-09-03
Cannot use GraphQLSchema from another module or realm


lal12:
**Describe the bug**
I installed express-graphql and typed-graphql, then I defined my entities and built the schema, when I pass the Schema to express-graphql and added it to my express app. As soon as I open grahpiql in my browser and run a query I get the error: 
```

{
  "errors": [
    {
      "message": "Cannot use GraphQLSchema \"[object Object]\" from another module or realm.\n\nEnsure that there is only one instance of \"graphql\" in the node_modules\ndirectory. If different versions of \"graphql\" are the dependencies of other\nrelied on modules, use \"resolutions\" to ensure only one version is installed.\n\nhttps://yarnpkg.com/en/docs/selective-version-resolutions\n\nDuplicate \"graphql\" modules cannot be used at the same time since different\nversions may have different capabilities and behavior. The data from one\nversion used in the function from another could produce confusing and\nspurious results."
    }
  ]
}
```

**Enviorment (please complete the following information):**
 - OS: Ubuntu 18.04
 - Node 10.9.0
 - Package version 0.13.1
 - TypeScript version 3.0.1

lal12:
I got it to work, by explicitly specifying the same version in my package.json as needed by type-graphql, but I guess it would be better if type-graphql would specify graphql as peerDependency, to prevent such problems.

19majkel94:
> But I guess it would be better if type-graphql would specify graphql as peerDependency, to prevent such problems.

It's not so simple. The main problem is that TypeGraphQL is tightly coupled to the version of `graphql-js`, so do the produced schema is. So I can't just put a wildcard `^0.12 || ^0.13 || >=14.0`.
If I make it a peerDependency, you still might have incompatibilities between TypeGraphQL and ecosystem, like 0.13 vs 14.0. Also it's error prone as you need to remember to install it too, even if you're not using `graphql-js` directly anywhere in your code.

capaj:
> So I can't just put a wildcard ^0.12 || ^0.13 || >=14.0.

So don't. Just specify the versions you know are compatible at the time of writing. Everyone does it like this. If a new version comes and it is compatible, you can add it later.

> Also it's error prone as you need to remember to install it too, even if you're not using graphql-js directly anywhere in your code.

I think you need to use it always-even apollo-server-core expects it at as peer dep.
```
Error: Cannot find module 'graphql'
    at Function.Module._resolveFilename (module.js:547:15)
[...]

PR #179, opened 2018-10-22
moved @types/graphql from dependencies to dev-dependencies

mechanical-turk:
image

As described in the FAQ page, multiple versions of @types/graphql causes build issues as the compiler gets confused. This is an attempt at decreasing the probability of having multiple versions of the package by moving @types/graphql to dev-dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -134,7 +134,7 @@
     },
     "@types/accepts": {
       "version": "1.3.5",
-      "resolved": "https://registry.npmjs.org/@types/accepts/-/accepts-1.3.5.tgz",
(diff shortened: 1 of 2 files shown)


mechanical-turk:
Tests passing:

image

19majkel94:
Unfortunatelly, we can't do that. It would cause that without manual installing of `@types/graphql` your app won't even compile, return type from `buildSchema` will be any, etc.

We could only move it do `peerDependency` and it would have to be done together with main `graphql` dependency but that was rejected a few weeks ago - see #148 and #144.

PR #190, opened 2018-11-06
Add Metadata Decorator [WIP] #124

DavidHooper:
Based on issues #124 and #163. Does this fit what you were thinking for the implementation?

Currently it only covers Fields + Objects until I understand you are happy with the simple implementation. From my understanding you want the Metadata decorator to be callable multiple times for the same objects/fields.

If this is what you're after, please point me in the right direction for writing some tests.


19majkel94:
Basically it looks ok. `Metadata` interface name might be confusing as we have `ParamMetadata` and other so it looks a bit like a base type.

In examples I would focus only on placing metadata, with fresh example, not copy-paste from the base, simple example. Copy-pasted mutations only introduce noise, not good use cases.

We should show all the possibilites of usage:
- placing metadata decorator once and twice
- creating own decorator for placing specific metadata key (e.g. the query-complexity case as it should be later refactored into metadata decorator usage)
- placing in:
  - object type
  - input type
  - query
  - fields
- reading the metadata (probably in `info` of resolver/middleware execution)
- others (maybe integration with some npm package for graphql?)

And of course the usage and all possibilities have to be well described in docs.

> If this is what you're after, please point me in the right direction for writing some tests.

I always try to test only the usage, so you should check if the metadata realy exist in all the types and fields after building the schema, if mutliple decorators overwrites the metadata key, etc.

You should create own test file for that, don't pollute the existing ones.

19majkel94:
@DavidHooper 
What is the progress of your work? Are you going to continue working on this feature? 😉

Issue #183, opened 2018-10-24
New schema generation pipeline


19majkel94:
Right now, schema generation pipeline works in this way:
- collect metadata from decorators (options, reflected type, placement (method/field, name, class)
- put them in raw shape in `MetadataStorage`
- "build" the metadata (by modifying `MetadataStorage`)
  - attach fields to types definition
  - attach args to resolvers definition
  - attach other metadata (middlewares, roles, etc.)
  - doing it sequentially, metadata type by type (fields, then handlers, etc.)
- generate the schema with `graphql-js` using "built" metadata from `MetadataStorage`
  - check for inheritance and interface implementing, then copy it's definitions
  - handle all overloading and overwriting cases

It wasn't designed for such complicated features as inheritance, etc., so it has started to became a bottleneck. So the schema generation pipeline has to be rewritten into a new design:
- `MetadataStorage` store all small parts of metadata from decorators (like now)
- `MetadataBuilder` takes `resolvers` and `types` to build the extended, full metadata
  - not modifying `MetadataStorage`, working like a "pure" function
  - doing it on demand (not ahead of time but lazily, step by step, recursive and with caching)
  - attaching all loose metadata to the base metadata definition
  - handling all extending, implementing and overwrtiting cases
- `SchemaGenerator` accepts built metadata object and generate schema corresponding to the metadata (without need for checking the metadata storage)

This would simplify this pipeline a lot, as well as allow to implement new features and allow to fix #133 and #110.

jdolle:
Something that may be worth considering during this refactor is multiple schemas in a single codebase. This is a requirement for my services because, for our application, there is a public graphql, which clients (web, mobile) interact with, and an internal graphql, with which other microservices interact. Placing these schemas in the same codebase allows easy code reuse.

If this is something you'd be open to, I'd be happy to work on it. I'm not familiar with this codebase, but there does not appear to be anything in the API that restricts it to only one schema.

19majkel94:
Of course, #110 is about separating schemas with a temporary solution for your codebase problem.

Evanion:
One of the things i reacted on was that you define a typeDef as a class in the current implementation.
I know it's probably because the interfaces gets stripped by `tsc` at compile time ...
but perhaps it would be possible to have a base interface like `Type`, `Enum`, `Scalar` etc.
and then create a ts transformer to convert interfaces that extend these base types, in to SDL format.

that way, it should be possible to write you typedefs more naturally (consider psudo code):
```
[...]

PR #203, opened 2018-11-22
support defaultValue for input types

benawad:
fix: https://github.com/19majkel94/type-graphql/issues/53
improved on: https://github.com/19majkel94/type-graphql/pull/141
I'm happy to iterate on this further if any changes are needed.

### DIFF
--- a/tests/functional/resolvers.ts
+++ b/tests/functional/resolvers.ts
@@ -44,6 +44,7 @@ import {
   PubSub,
   PubSubEngine,
   ClassType,
+  ConflictingDefaultValuesError,
 } from "../../src";
 import { getMetadataStorage } from "../../src/metadata/getMetadataStorage";
 import { IOCContainer } from "../../src/utils/container";
@@ -464,7 +465,7 @@ describe("Resolvers", () => {
       });
 
       it("should throw error when defaultValue in decorator doesn't match implicit defaultValue in input type", async () => {
-        expect.assertions(2);
+        expect.assertions(7);
 
         @InputType()
         class InputWithDefaultField {
@@ -485,8 +486,13 @@ describe("Resolvers", () => {
           });
         } catch (err) {
           expect(err).toBeInstanceOf(Error);
-          const error = err as Error;
+          expect(err).toBeInstanceOf(ConflictingDefaultValuesError);
+          const error = err as ConflictingDefaultValuesError;
           expect(error.message).toContain("conflicting default values");
+          expect(error.message).toContain("InputWithDefaultField");
+          expect(error.message).toContain("twoDefaultValuesField");
+          expect(error.message).toContain("implicitDefaultValue");
+          expect(error.message).toContain("explicitDefaultValue");
         }
       });
     });
--- a/tests/functional/resolvers.ts
+++ b/tests/functional/resolvers.ts
@@ -16,6 +16,7 @@ import {
   ValidationContext,
   visit,
   visitWithTypeInfo,
+  IntrospectionInputObjectType,
 } from "graphql";
 import * as path from "path";
 import { plainToClass } from "class-transformer";
@@ -56,6 +57,7 @@ describe("Resolvers", () => {
     let mutationType: IntrospectionObjectType;
     let sampleObjectType: IntrospectionObjectType;
     let argMethodField: IntrospectionField;
+    let sampleInputType: IntrospectionInputObjectType;
 
     beforeAll(async () => {
       getMetadataStorage().clear();
@@ -64,6 +66,8 @@ describe("Resolvers", () => {
       class SampleInput {
         @Field()
         field: string;
+        @Field({ defaultValue: "defaultStringFieldDefaultValue" })
+        defaultStringField?: string;
       }
 
       @ArgsType()
@@ -228,7 +232,9 @@ describe("Resolvers", () => {
         type => type.name === "SampleObject",
       ) as IntrospectionObjectType;
       argMethodField = sampleObjectType.fields.find(field => field.name === "argMethodField")!;
-      console.log(schemaIntrospection.types);
+      sampleInputType = schemaIntrospection.types.find(
+        field => field.name === "SampleInput",
+      )! as IntrospectionInputObjectType;
     });
 
     // helpers
@@ -388,6 +394,19 @@ describe("Resolvers", () => {
         expect(inputArg.defaultValue).toBe('"defaultStringArgDefaultValue"');
         expect(nullableDefaultValueStringArgType.kind).toEqual(TypeKind.SCALAR);
         expect(nullableDefaultValueStringArgType.name).toEqual("String");
+      });
+    });
+
+    describe("Input object", () => {
+      it("should generate nullable string arg type with defaultValue for input object field", async () => {
+        const defaultValueStringField = sampleInputType.inputFields.find(
+          arg => arg.name === "defaultStringField",
+        )!;
+        const defaultValueStringFieldType = defaultValueStringField.type as IntrospectionNamedTypeRef;
+
+        expect(defaultValueStringField.defaultValue).toBe('"defaultStringFieldDefaultValue"');
+        expect(defaultValueStringFieldType.kind).toEqual(TypeKind.SCALAR);
+        expect(defaultValueStringFieldType.name).toEqual("String");
       });
     });
 
--- a/tests/functional/resolvers.ts
+++ b/tests/functional/resolvers.ts
@@ -495,38 +495,6 @@ describe("Resolvers", () => {
         expect(inheritDefaultFieldType.kind).toEqual(TypeKind.SCALAR);
         expect(inheritDefaultFieldType.name).toEqual("String");
       });
-
-      it("should throw error when defaultValue in decorator doesn't match implicit defaultValue in input type", async () => {
-        expect.assertions(7);
-
-        @InputType()
-        class InputWithDefaultField {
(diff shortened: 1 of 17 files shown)


19majkel94:
Looks ok but it's just a beginning. 
Please take a look at this comment:
https://github.com/19majkel94/type-graphql/pull/141#pullrequestreview-151603632

tests/functional/resolvers.ts
```diff
@@ -104,6 +104,7 @@ describe("Resolvers", () => {
           @Arg("stringArrayArg", type => String) stringArrayArg: string[],
           @Arg("explicitArrayArg", type => [String]) explicitArrayArg: any,
           @Arg("nullableStringArg", { nullable: true }) nullableStringArg?: string,
+          @Arg("defaultStringArg", { defaultValue: "bob" }) defaultStringArg?: string,
```
19majkel94: Please try to follow the descriptive names convention, like `defaultStringArgDefaultValue` instead of `foo`, `bar`, `bob` and other sample names that might generate conflicts.

benawad:
What is `Support for initializers in classes ("reflection") - creating an instance of the class and checking the value of the properties to detect default values`?

19majkel94:
@benawad 
```ts
@InputType()
class SampleInput {
  @Field()
  defaultField: string = "defaultFieldValue"
}
```
```graphql
input SampleInput {
  defaultField: String = "defaultFieldValue"
}
```

benawad:
How should I go about doing that?

19majkel94:
@benawad 
Just by "creating an instance of the class and checking the value of the properties to detect default values" 😃 

```ts
class SampleInput {
  normalField: string;
  stringFieldWithDefault: string = "defaultValue"
}

const instance = new SampleInput();

const defaultsMap = Object.keys(instance).map(field => ({
  field,
  defaultValue: instance[field],
}));

console.log(defaultsMap);
```

benawad:
I was able to get it working for input objects, but I'm having trouble figuring out how to do it with args

19majkel94:
@benawad 
You need to do the same things as with input - just create the instance in `mapArgFields`, read the defaults and attach them to `args[field.schemaName]` config:
```ts
const instance = new (argumentType.target as any)();
```

tests/functional/resolvers.ts
```diff
@@ -104,6 +114,10 @@ describe("Resolvers", () => {
           @Arg("stringArrayArg", type => String) stringArrayArg: string[],
           @Arg("explicitArrayArg", type => [String]) explicitArrayArg: any,
           @Arg("nullableStringArg", { nullable: true }) nullableStringArg?: string,
+          @Arg("defaultStringArg", { defaultValue: "defaultStringArgDefaultValue" })
```
19majkel94: Object types have no default values, only input types can have by GraphQL specification

19majkel94:
We should also test some edge cases, like overwriting default values in child classes (implicit and by decorator option):
```ts
class Base {
  prop = "baseProp"
}

class Child extends Base {
  prop = "childProp"
}
```

And I think that we should report to user when `implicitDefaultValue` has the value and `field.typeOptions.defaultValue` has the value and they're not equal to each other.
Now it's just silently overwrites the default value from decorator.

benawad:
Is it possible to detect the default values for function parameters?
```
          @Arg("implicitDefaultStringArg")
          implicitDefaultStringArg: string = "implicitDefaultStringArgDefaultValue",
```

benawad:
Would you like me to make any other changes?

tests/functional/resolvers.ts
```diff
@@ -224,6 +246,12 @@ describe("Resolvers", () => {
         type => type.name === "SampleObject",
       ) as IntrospectionObjectType;
       argMethodField = sampleObjectType.fields.find(field => field.name === "argMethodField")!;
+      sampleInputType = schemaIntrospection.types.find(
```
19majkel94: This variables are not used globally in this test suite - move them do the used block:
```ts
describe("Input object", () => {
  let sampleInputType: IntrospectionInputObjectType;
  let sampleInputChildType: IntrospectionInputObjectType;

  beforeAll(() => {
    sampleInputType = schemaIntrospection.types.find(
      field => field.name === "SampleInput",
    )! as IntrospectionInputObjectType;
    sampleInputChildType = schemaIntrospection.types.find(
      field => field.name === "SampleInputChild",
    )! as IntrospectionInputObjectType;
  });
```

tests/functional/resolvers.ts
```diff
@@ -100,10 +119,13 @@ describe("Resolvers", () => {
           @Arg("booleanArg") booleanArg: boolean,
           @Arg("numberArg") numberArg: number,
           @Arg("inputArg") inputArg: SampleInput,
+          @Arg("inputChildArg") inputChildArg: SampleInputChild,
           @Arg("explicitNullableArg", type => String, { nullable: true }) explicitNullableArg: any,
           @Arg("stringArrayArg", type => String) stringArrayArg: string[],
           @Arg("explicitArrayArg", type => [String]) explicitArrayArg: any,
           @Arg("nullableStringArg", { nullable: true }) nullableStringArg?: string,
+          @Arg("defaultStringArg", { defaultValue: "defaultStringArgDefaultValue" })
+          defaultStringArg?: string,
```
19majkel94: It has `defaultValue` so it's not optional (`?`) anymore

tests/functional/resolvers.ts
```diff
@@ -74,6 +89,10 @@ describe("Resolvers", () => {
         numberArg: number;
         @Field()
         inputObjectArg: SampleInput;
+        @Field({ defaultValue: "defaultStringArgDefaultValue" })
+        defaultStringArg?: string;
```
19majkel94: The same - no optional when defaults used

tests/functional/resolvers.ts
```diff
@@ -74,6 +89,10 @@ describe("Resolvers", () => {
...
+        @Field()
+        implicitDefaultStringArg?: string = "implicitDefaultStringArgDefaultValue";
```
19majkel94: No optional when defaults used

tests/functional/resolvers.ts
```diff
@@ -375,6 +403,92 @@ describe("Resolvers", () => {
         expect(inputArgInnerType.kind).toEqual(TypeKind.INPUT_OBJECT);
         expect(inputArgInnerType.name).toEqual("SampleInput");
       });
+
+      it("should generate nullable string arg type with defaultValue for object field method", async () => {
+        const inputArg = argMethodField.args.find(arg => arg.name === "defaultStringArg")!;
+        const defaultValueStringArgType = inputArg.type as IntrospectionNamedTypeRef;
+
+        expect(inputArg.defaultValue).toBe('"defaultStringArgDefaultValue"');
```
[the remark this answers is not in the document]
19majkel94: Ok, it looks like it just works like that. Weird:
```
        {
          "name": "ratingsCount",
          "args": [
            {
              "name": "minRate",
              "defaultValue": "0"
            }
          ]
        },
```
So it has to stay like this 😕

19majkel94:
@benawad 
> Would you like me to make any other changes?

Yes, we need to add info about all this things in docs 😃 
https://19majkel94.github.io/type-graphql/docs/resolvers.html
- `defaultValue` in `@Arg` decorator
- `defaultValue` in `@Field` decorator (only for `@InputType` and `@ArgsType`)
- initialization of property affects default value in schema
- inheritance of default values
- warning about conflict between decorator and and initializer default value

Also, examples and docs might need some updates when there's a default value together with `{ nullable: true }` decorator's option.

And how do we handle the edge case when parent class has property initializer but child try to overwrites it with decorator option? Does it throw error?
And what about when parent has decorator option and child overwrites it with property initalizer?
Do we even test the simple inheritance with no overwrites? 😕

benawad:
"And how do we handle the edge case when parent class has property initializer but child try to overwrites it with decorator option? Does it throw error?"
"And what about when parent has decorator option and child overwrites it with property initalizer?"

I think we should take the child defaultValue in both cases without throwing an error because they could be inheriting from a class from a library or something and not know (and don't care) how thaat class set the default value. 

"Do we even test the simple inheritance with no overwrites? 😕"

Nope, and I just added a test for it and it failed 😂. Where in the code do you handle inheritance?

I'll start working on the docs next.

19majkel94:
> Where in the code do you handle inheritance?

Before #183 it's all over the place 😆
Now I've realized that `@Authorized` also doesn't work with inheritance 😞 	

Args classes inheritance is here:
https://github.com/benawad/type-graphql/blob/d7659fb0d446d3a35ba7943b7f82a60bb939d3bb/src/schema/schema-generator.ts#L453-L460

Input types inheritance is here:
https://github.com/benawad/type-graphql/blob/d7659fb0d446d3a35ba7943b7f82a60bb939d3bb/src/schema/schema-generator.ts#L330-L336

> I think we should take the child defaultValue in both cases

Right! Make sure that the tests are covering this cases 😉 	

Thanks and keep going! 💪

19majkel94:
@benawad
> start working on docs

Are you going to finish this up or should I checkout your branch and add some changes there?

benawad:
I'm not sure when I'll get time to finish this up, so it would be great if you want to.

19majkel94:
@benawad 
I think that we are ready to go with this feature.
Can you take a look at my changes and review them? 😉

benawad:
I'll take a look at it tonight

benawad:
Looks good

PR #210, opened 2018-12-05
use Object.create instead of new to avoid constructor side effects

domi7777:
I've encountered a problem with a resolver, when code is invoked to create an instance of a class. Unfortunately, the constructor of my class is expecting a non-optional argument, and would fail without it.

Therefore I recommend creating the instance of the class without invoking the constructor, using ```Object.create(Type.prototype)```

see also https://stackoverflow.com/a/50856428

### DIFF
--- a/src/helpers/types.ts
+++ b/src/helpers/types.ts
@@ -65,7 +65,7 @@ export function convertToType(Target: any, data?: object): object | undefined {
     return data;
   }
 
-  return Object.assign(new Target(), data);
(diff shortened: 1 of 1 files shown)


19majkel94:
Unfortunately, without invoking the constructor, field defaults doesn't work, as you can see in the broken tests.

Also, putting logic in constructor is strongly prohibited - your graphql types should be simple DTOs, not rich models.

So I have to reject this PR 🔒

Issue #119, opened 2018-07-28
Support for lists with nullable elements


erencay:
AFAIK there is currently no support for lists with nullable elements, such as `[String]!`.

```
export function wrapWithTypeOptions<T extends GraphQLType>(
  type: T,
  typeOptions: TypeOptions = {},
): T {
  let gqlType: GraphQLType = type;
  if (typeOptions.array) {
    gqlType = new GraphQLList(new GraphQLNonNull(gqlType));
  }
  if (!typeOptions.nullable) {
    gqlType = new GraphQLNonNull(gqlType);
  }
  return gqlType as T;
}
```
Seems like this method forces all lists to have non-null elements.

Appopriate solutions would be either adding a null value to value array 
```@Field(type => [String, null])``` 

Or providing a decorator option such as;
```@Field(type => [String], { nullable: true, nullableElement: true })```

@19majkel94 would you like me to make a pull request?

**Comment by erencay** [comment]
In my case I'm building a game and I have a fixed length array which filled with null values by default. As the game progress I need to mutate values at certain indexes. So I think one might need this feature where there is a fixed length of array filled with null values, index matters and array contains non-scalar type.

Something like:
```
[null, null, null, null, null, null, null, null, null]
[null, null, null, null, null, null, { number: 1, color: "red" }, null, null]
```
There are workarounds but this structure feels most natural.

**Comment by 19majkel94** [comment]
I think that your app should abstract you internal, array representation from the public API.

In GraphQL there's no array, only lists (`GraphQLList` as you see above), which means you cant access random element of the list.

 So you should use the fieldResolver pattern to transform an array to list of entries:
```gql
type GameArrayItem {
  index: Int!
  value: GameItemValue!
}

type GameItemValue {
  number: Int!
  color: String!
}
```
So you query should return list of `GameArrayItem`:
```gql
type Query {
  items: [GameArrayItem!]!
  item(index: Int!): GameArrayItem
}

query GetItems {
  items {
    index
    value {
      number
      color
    }
}

query GetItem1 {
  item(index: 1) {
    value {
      number
      color
    }
}
```

And mutation should operate on `index`:
```gql
type Mutation {
  gameProgress(index: Int!): Boolean
}

input GameProgressInput {
  number: Int!
  color: String!
}

mutation ProgressGame {
  gameProgress({
	index: 1
	number: 55
    color: "blue"
  })
}
```
Not on the whole list replacing items by itself.

erencay:
[...]

PR #211, opened 2018-12-06
Add more options for array with nullable element #119

kenpusney:
This is to add new options for `nullable` to resolve issue #119


src/helpers/types.ts
```diff
@@ -42,7 +42,13 @@ export function wrapWithTypeOptions<T extends GraphQLType>(
 ): T {
   let gqlType: GraphQLType = type;
   if (typeOptions.array) {
-    gqlType = new GraphQLList(new GraphQLNonNull(gqlType));
+    if (typeOptions.nullable === "item") {
```
19majkel94: A bit too complicated - please change this function body to:
```ts
  let gqlType: GraphQLType = type;
  if (typeOptions.array) {
    if (typeOptions.nullable === "item" || typeOptions.nullable === "itemAndList") {
      gqlType = new GraphQLList(gqlType);
    } else {
      gqlType = new GraphQLList(new GraphQLNonNull(gqlType));
    }
  }
  if (!typeOptions.nullable || typeOptions.nullable === "item") {
    gqlType = new GraphQLNonNull(gqlType);
  }
  return gqlType as T;
```

tests/functional/resolvers.ts
```diff
@@ -544,6 +554,29 @@ describe("Resolvers", () => {
         expect(itemType.name).toEqual("String");
       });
 
+      it("should generate explicit array of nullable string return type for query", async () => {
+        const explicitNullableItemArrayQuery = getQuery("explicitNullableItemArrayQuery");
+        const type = explicitNullableItemArrayQuery.type as IntrospectionNonNullTypeRef;
+        const listType = type.ofType as IntrospectionListTypeRef;
+        const itemType = listType.ofType as IntrospectionNamedTypeRef;
+
+        expect(listType.kind).toEqual(TypeKind.LIST);
```
19majkel94: Please check also the `type`, like for fields:
`expect(nonNullArrayType.kind).toEqual(TypeKind.NON_NULL);`

src/helpers/types.ts
kenpusney: I don't think this will make it simpler since there may have some weird behavior when a non-array field annotated as `@Field( nullalbe = 'items' )`. it would be better just limit those options only for array.

19majkel94: We should just throw an error to the user when a non-array field is annotated as `nullable: "items"`. Silently converting `items` or `itemsAndList` to `true` (which is happening right now) is not a good idea.

19majkel94:
@kenpusney 
Are you going to finish this up or should I checkout your branch and add some changes there?
We have to also add a proper info about this options in docs 📝

kenpusney:
@19majkel94 
sorry get lost in massive work, possibly have no enough time to finish this. Please do help if you can, thanks.

19majkel94:
@kenpusney 
Once again: thank you very much for your contribution! 😉

PR #212, opened 2018-12-08
Make Golang a serious language in docs

MichalLytek:


pkosiec:
tenor-213008119

Good. Go is a serious language indeed. It has the same `enum` functionality known from C, just without `enum` keyword.

PR #216, opened 2018-12-22
Fix browser-shim.ts

samdenty:
Returning undefined is insufficient as not all polyfills for decorator support check for null

### DIFF
--- a/src/browser-shim.ts
+++ b/src/browser-shim.ts
@@ -12,7 +12,7 @@ plugins: [
 ]
 */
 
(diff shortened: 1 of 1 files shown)


src/browser-shim.ts
```diff
@@ -12,7 +12,7 @@ plugins: [
 ]
 */
 
-const dummyFn = () => void 0;
+const dummyFn = () => () => void 0;
```
19majkel94: Can you change the name to e.g. `dummyDecorator`?

PR #237, opened 2019-01-21
Circular dependencies fix

davewasmer:
Tweaks the implementation of `findType` so that the return type function is called lazily rather than synchronously. This allows users to define a lazily-loaded type in that type function, and not create problematic circular dependency loops:

```ts
@Field(_type => require('./some/other/class'))
       // ^ this function won't be invoked until after the module finishes loading
```


19majkel94:
@davewasmer 
Thanks for showing your rare use case and fixing this up! 🚀 

Side note: To make your case work you need to switch to `emitDecoratorMetadata: false` and provide all types by decorators or switch to `any`.

Maybe in the future the reflection in circular cases will be fixed:
https://github.com/Microsoft/TypeScript/issues/14971

XBeg9:
@19majkel94 do you plan to release this in near future?

19majkel94:
@XBeg9 
`0.17.0-beta.4` - `npm i type-graphql@next`

PR #242, opened 2019-01-27
fix: special root fields don't have __isTypeOf

birkir:
I have a case where two different graphql parsers fail because of the root fields (in @graphql-modules)

Should this property really be exposed at all?

### DIFF
--- a/src/utils/createResolversMap.ts
+++ b/src/utils/createResolversMap.ts
@@ -23,6 +23,10 @@ export function createResolversMap(schema: GraphQLSchema): ResolversMap {
           __isTypeOf: type.isTypeOf || undefined,
           ...generateFieldsResolvers(type.getFields()),
         };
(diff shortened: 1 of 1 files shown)


19majkel94:
@birkir 
Could you elaborate a little bit more about your use case, where the `undefined` value brokes things? 😕 I would like to create a test case first, before fixing unknown bug.

birkir:
Thanks for the quick reply and fix. I was working late, so I forgot to add tests and explain this in more detail.

Actually what you ended up doing is way better, because as it turns out, traversers also fail on non root types.

But the test you made is still valid.

I was having an issue where a traverser was expecting every property on a resolver as an object, so in my case I had the error "Cannot use 'in' operator to search for 'subscribe' in undefined" (where undefined being the one you just eliminated)

PR #243, opened 2019-01-29
[WIP] fix(resolvers): adding custom decorators doesn't break

### LINKED ISSUE #166: type-graphql decorators do not play nice with other decorators
**Describe the bug**
When using type-graphql decorators (`Query`, `Mutation`, etc.), other decorators applied to the same method get nullified and no longer execute at run time. 

**To Reproduce**
```ts
export function LogMe(): MethodDecorator {
  return (obj, methodName, descriptor: any) => {
    const originalMethod: Function = descriptor.value
    descriptor.value = function() {
      console.log(`${methodName.toString()} exeucting`)
      const res = originalMethod.apply(this, arguments)
      return res
    } 
  }
}

@Resolver()
export class MyResolver {
  @Query(() => String)
  @LogMe()
  async helloWorld() {
    return 'Hello World'
  }
}
```
**Expected behavior**
Expected to see `helloWorld executing` being output'ed to the console when running GraphQL query. When in fact nothing shows up

 - OS: macOS 10.13.3 (17D47)
 - Node v10.6.0
 - Package version 0.14.0
 - TypeScript version 3.1.3

**Additional context**
I believe this is due to the way `getHandlerInfo` inside `handlers.ts` is implemented. in particular

```ts
   handler: (prototype as any)[methodName],
```

My guess is that `prototype[methodName]` contains the original method without any decorations, which is why the `LogMe` decorator` is no longer executing. 

Why not use the `value` field of the `descriptor` third argument provided to a method decorator instead?

ematipico:
Fixes #166.

Although I would like to add a test but I don't know what I should test. Suggestions?

### DIFF
--- a/src/resolvers/create.ts
+++ b/src/resolvers/create.ts
@@ -35,7 +35,7 @@ export function createHandlerResolver(
         globalValidate,
         pubSub,
       );
-      return resolverMetadata.handler!.apply(targetInstance, params);
+      return targetInstance[resolverMetadata.methodName].apply(targetInstance, params);
     });
   };
 }
--- a/tests/functional/resolvers.ts
+++ b/tests/functional/resolvers.ts
@@ -61,6 +61,16 @@ describe("Resolvers", () => {
     let sampleObjectType: IntrospectionObjectType;
     let argMethodField: IntrospectionField;
 
+    function LogMe(): MethodDecorator {
+      return (obj, methodName, descriptor: any) => {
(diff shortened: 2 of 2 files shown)


19majkel94:
> I don't know what I should test

You should create a failing test first, that describes the bug or the issue.
Then you can dig in and fix the code until all tests passes. Then you can refactor the code 😉 

You should also remove collecting the `handler` in metadata storage as it's not used anymore.

PR #244, opened 2019-01-29
[WIP] Generic Type

OwenCalvin:
Now you can create Models to make "Generics" GraphQL classes, the injected classes with be into @Destination properties


OwenCalvin:
[WIP] This is not fully finished yet but it's usable

OwenCalvin:
> @OwenCalvin
> I still don't know what this feature is all about.
> 
> Please describe your proposal in more details: background about use cases, the limitation of the current API, the proposed solution with API draft.

Yes I'm working on it... I'll post it soon

OwenCalvin:
@19majkel94 So for the moment, I've made an example into examples/generic-type, with this you can see what is the goal if the feature

OwenCalvin:
The limitation of the current API:


 the proposed solution with API draft.

OwenCalvin:
### Use Case:
Let suppose that you have multiple ObjectType, a QueryResponse class that is generic and has a dynamic type field :

**My ObjectTypes:**  
image  
image  

**My QueryResponse generic class that has a generic property _items_ :**  
image   
    
  
   
So now if you want to return a QueryResponse in a cars query (QueryResponse<Car>) and the same for a persons query (QueryResponse<Person>). So you do not need to redeclare a new class to type staticly items field.    

You can use the new type-graphql generic typing feature in queries like this (Also usable as a ArgsType or InputType: see examples/generic-type):  
image

OwenCalvin:
I closed the pull request to create a new cleaned one, you can delete it

Issue #180, opened 2018-10-22
Generic types support


19majkel94:
Right now we can only extend other object type, input type and args type classes. But this doesn't allow us to overwrite a property type so we can't create a truly generic types like the one used for pagination.

So, to the higher order class pattern we have to introduce `isAbstract` option that works the same way as in [resolvers inheritance](https://19majkel94.github.io/type-graphql/docs/interfaces-and-inheritance.html#resolvers-inheritance). Then we can create generic types:
```ts
export function Edge<TNodeType>(NodeType: ClassType<TNodeType>) {
  @ObjectType({ isAbstract: true })
  abstract class EdgeType {
    @Field(type => NodeType)
    node: TNodeType;
    @Field()
    cursor: string;
  }
  return EdgeType;
}
```

We can then easily extend the generic class:
```ts
@ObjectType()
export class RecipeEdge extends Edge(Recipe) {
  @Field()
  personalNotes: string;
}

@ObjectType()
export class FriendshipEdge extends Edge(User) {
  @Field()
  friendedAt: Date;
}
```

So this would generate following GraphQL type:
```graphql
type RecipeEdge {
  node: Recipe!
  cursor: String!
  personalNotes: String!
}

type FriendshipEdge {
  node: User!
  cursor: String!
  friendedAt: DateTime!
}
```

OwenCalvin:
@19majkel94 Any progress ?

19majkel94:
Available in `0.17.0-beta.8` 🚀

XBeg9:
Amazing news ;) thanks @19majkel94

Veetaha:
@19majkel94 According to your example:

> `` @ObjectType({ name: `Paginated${TItemClass.name}Response` })``

This way you propose to generate unique names for generic types. But what if I specified another name for `TItemClass` when had decorated it with `@ObjectType('SomeOtherName')`? I would like to set the name of the generated class without having to repeat that `'SomeOtherName'` as a factory function parameter. Could you please expose a method to retrieve GraphQL name from the given class constructor that was decorated with `@ObjectType()` ?

19majkel94:
I'm not sure what you are referring to 😕 

You can have a function that generates a classes (no `isAbstract`, use `name` for unique type names):

```ts
function Foo(TItem) {
  @ObjectType({ name: "Foo" + TItem.name })
  class Foo {
    @Field()
	bar: string;
  }
  return Foo;
}

const FooBar = Foo(Bar);
[...]

PR #245, opened 2019-01-31
Generic Type

OwenCalvin:
### Use Case:
I've made an example into examples/generic-type, with this you can see what is the goal of the feature.  
Let suppose that you have multiple ObjectType, a QueryResponse class that is generic and has a dynamic type field :

**My ObjectTypes:**  
image  
image  

**My QueryResponse generic class that has a generic property _items_ :**  
image   
    
  
   
So now if you want to return a QueryResponse in a cars query (QueryResponse<Car>) and the same for a persons query (QueryResponse<Person>). So you do not need to redeclare a new class to type staticly items field.    

You can use the new type-graphql generic typing feature in queries like this (Also usable as a ArgsType or InputType: see examples/generic-type):  
image
  
  
  
**Compiled queries**
image
image


19majkel94:
@OwenCalvin 
I really appreciate your work, looks like you've spend a lot of time 💪 

But I wasn't asking for help with generics - as you can see in #180, there's no "Help wanted" badge and I'm assigned to this issue to do this by my own.

The proposed API is definitely not the one I would like to introduce. It's too complicated, missleading (`type =>` and now `{ type: }` too), there's too much magic going behind and not enough typesafe.

I have different design goals and feature implementations/bug solution in my head, so I can't accept your PR.

I've placed and info in the readme long, long time ago to prevent this kind of situation:
> If you want to add a new big feature, please create a proposal first, where we can discuss the idea and implementation details. This will prevent wasting of your time if the PR be rejected.

But feel free to continue your work in your fork `@owencalvin/type-graphql` or maybe as a different framework (like @goldcaddy77 [warthog](https://github.com/goldcaddy77/warthog) if you need this kind on feature and you like this API.

OwenCalvin:
Okay no problem, I needed this so I made it :)
I will continue my work but I keep an eye on you implementation :)

PR #249, opened 2019-02-02
skip superfluous type conversion

### LINKED ISSUE #229: Skip superfluous type conversion
`convertToType` helper from `helpers/types.ts` will convert type, even if `data` is already of `Target` type.

https://github.com/19majkel94/type-graphql/blob/master/src/helpers/types.ts#L80

Could be avoided by adding
```ts
  if (data instanceof Target) {
    return data;
  }
```
and potentially fix issues when using any kind of logic in object type constructors.

Agreon:
This PR fixes Constructor-Side-Effects that can occur when a type is converted even though it is not necessary.
Closes #229

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ## Fixes
 - fix calling return type getter function `@Field(type => Foo)` before finishing module evaluation (allow for extending circular classes using `require`)
 - fix nullifying other custom method decorators - call the method on target instance, not the stored reference to original function (#247)
+- skip superfluous type-conversion. (#229)
 
 ## v0.16.0
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 ## Fixes
 - fix calling return type getter function `@Field(type => Foo)` before finishing module evaluation (allow for extending circular classes using `require`)
 - fix nullifying other custom method decorators - call the method on target instance, not the stored reference to original function (#247)
-- skip superfluous type-conversion. (#229)
+- Prevent unnecessary conversion of a object that is already of the requested typ, to avoid constructor side-effects (#229)
 
 ## v0.16.0
 ### Features
--- a/src/helpers/types.ts
+++ b/src/helpers/types.ts
@@ -83,7 +83,7 @@ export function convertToType(Target: any, data?: object): object | undefined {
     return;
   }
   // Prevent unecessary conversion
-  if (data instanceof Target) {
+  if (typeof Target === "object" || data instanceof Target) {
     return data;
   }
   // skip converting scalars (object scalar mostly)
--- a/src/helpers/types.ts
+++ b/src/helpers/types.ts
@@ -82,6 +82,10 @@ export function convertToType(Target: any, data?: object): object | undefined {
   if (data == null) {
     return;
   }
+  // Prevent unecessary conversion
+  if (data instanceof Target) {
+    return data;
+  }
   // skip converting scalars (object scalar mostly)
   if (Target instanceof GraphQLScalarType) {
     return data;
--- a/tests/units/convertToType.ts
+++ b/tests/units/convertToType.ts
@@ -1,0 +1,35 @@
+import { convertToType } from "../../src/helpers/types";
+
+describe("convertToType", () => {
+  let globalCounter = 0;
+
+  class SampleTarget {
+    constructor(public test: number) {
+      globalCounter++;
+    }
+  }
+
+  beforeEach(() => {
+    globalCounter = 0;
+  });
+
+  it("should properly convert a object to the defined target", async () => {
+    const testObject = {
+      test: 4,
+    };
+
+    const convertedObject = convertToType(SampleTarget, testObject);
+    expect(convertedObject).toBeDefined();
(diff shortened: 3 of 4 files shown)


tests/units/convertToType.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  it("should properly convert a object to the defined target", async () => {
+    const testObject = {
+      test: 4,
+    };
+
+    const convertedObject = convertToType(SampleTarget, testObject);
+    expect(convertedObject).toBeDefined();
+    expect(convertedObject).toHaveProperty("test");
+    expect(globalCounter).toBe(1);
```
19majkel94: Missing check for `instanceOf` - for now it can just return the `data` and the test passes.
We have to check if it really converts the object to a class instance.

tests/units/convertToType.ts
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  it("should properly convert a object to the defined target", async () => {
+    const testObject = {
+      test: 4,
+    };
+
+    const convertedObject = convertToType(SampleTarget, testObject);
+    expect(convertedObject).toBeDefined();
+    expect(convertedObject).toHaveProperty("test");
```
19majkel94: Let's check also the value

tests/units/convertToType.ts
```diff
@@ -0,0 +1,35 @@
...
+    expect(globalCounter).toBe(1);
+  });
+
+  it("should not convert a object to the defined target if it is already instance of that Target", async () => {
+    const testObject = new SampleTarget(4);
+
+    const convertedObject = convertToType(SampleTarget, testObject);
+    expect(convertedObject).toBeDefined();
+    expect(convertedObject).toHaveProperty("test");
+    expect(globalCounter).toBe(1);
```
19majkel94: I would add this check also in line 29 - before converting. It makes the test more readable, easier to understand what behavior we're defining, rather that remembering the side effect of the constructor.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 ## Fixes
 - fix calling return type getter function `@Field(type => Foo)` before finishing module evaluation (allow for extending circular classes using `require`)
 - fix nullifying other custom method decorators - call the method on target instance, not the stored reference to original function (#247)
+- skip superfluous type-conversion. (#229)
```
19majkel94: - No `.` at the end.
- Let's make it more developer friendly message - "Prevent unnecessary conversion of object that are instance of object type class" or something like that. Write what you would like to read in the changelog after installing new TypeGraphQL version from npm.

Agreon:
The tests failed, because an object was given to the convert-method as Target and instanceof was throwing an Error in this case. Fixed it ;)

src/helpers/types.ts
```diff
@@ -83,7 +83,7 @@ export function convertToType(Target: any, data?: object): object | undefined {
     return;
   }
   // Prevent unecessary conversion
-  if (data instanceof Target) {
+  if (typeof Target === "object" || data instanceof Target) {
```
19majkel94: Maybe it's better to have just `if (data instanceof Target)` but placed in line 97?
Then all the scalar and simple types cases will be handled.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 ## Fixes
 - fix calling return type getter function `@Field(type => Foo)` before finishing module evaluation (allow for extending circular classes using `require`)
 - fix nullifying other custom method decorators - call the method on target instance, not the stored reference to original function (#247)
+- Prevent unnecessary conversion of a object that is already of the requested typ, to avoid constructor side-effects (#229)
```
19majkel94: I usually refer to pull requests, not the original issues - so this should be #249.

tests/units/convertToType.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  it("should properly convert a object to the defined target", async () => {
+    const testObject = {
+      test: 4,
+    };
+
+    expect(globalCounter).toBe(0);
+    const convertedObject = convertToType(SampleTarget, testObject);
+    expect(convertedObject).toBeDefined();
+    expect(convertedObject instanceof SampleTarget).toBe(true);
```
19majkel94:
```suggestion
    expect(convertedObject).toBeInstanceOf(SampleTarget);
```

tests/units/convertToType.ts
```diff
@@ -0,0 +1,41 @@
...
+    expect(globalCounter).toBe(1);
+  });
+
+  it("should not convert a object to the defined target if it is already instance of that Target", async () => {
+    const testObject = new SampleTarget(4);
+
+    expect(globalCounter).toBe(1);
+    const convertedObject = convertToType(SampleTarget, testObject);
+    expect(convertedObject).toBeDefined();
+    expect(convertedObject instanceof SampleTarget).toBe(true);
```
19majkel94:
```suggestion
    expect(convertedObject).toBeInstanceOf(SampleTarget);
```

Agreon:
Unfortunately i cannot reproduce the problem with a functional test.

19majkel94:
Thanks anyway for your contribution! ❤️ 

I've fixed it via 7bddb04 🔒

PR #255, opened 2019-02-16
Generic types support

### LINKED ISSUE #180: Generic types support

19majkel94:
This PR closes #180 🔒


frankdugan3:
This feature looks awesome! Any chance for a release bump that includes this feature?

19majkel94:
@frankdugan3 See https://github.com/19majkel94/type-graphql/issues/180#issuecomment-466171745

Issue #266, opened 2019-03-01
emitSchemaFile: Create directory automatic


mszekiel:
Using `emitSchemaFile` from `buildSchema` I want to show directory in which `.gql` file will be stored. But target location is not being created, I have to create it by myself.  I'd like to use it simple by passing directory only, with automatic directory creation.

xanaxx:
@19majkel94 what's the node version requirement? Node `v10.12.0` implements recursive mkdir so `emitSchemaDefinitionFile` functions could be clean and stay native.

19majkel94:
@xanaxx 
https://github.com/19majkel94/type-graphql/blob/40d86913ba30411d000d02b27d4f3c7c16463786/.travis.yml#L5-L9

I will drop v6 in the end of April 2019.

19majkel94:
Closed by #269 🔒

PR #269, opened 2019-03-03
feat(emit-schema): add support for recursive directory creation

mszekiel:
#266


src/utils/emitSchemaDefinitionFile.ts
```diff
@@ -15,12 +15,26 @@ const generatedSchemaWarning = /* graphql */ `\
...
 `;
 
+export function mkdirRecursive(targetPath: string) {
+  path
+    .parse(path.resolve(targetPath))
+    .dir.split(path.sep)
+    .reduce((previousPath, folder) => {
+      const currentPath = path.join(previousPath, folder, path.sep);
+      if (!existsSync(currentPath)) {
+        mkdirSync(currentPath);
```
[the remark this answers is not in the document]
19majkel94: Maybe create a function that parse the path to an array of sequential directories to create and then just call it in sync and async versions of the `mkdirRecursive` function.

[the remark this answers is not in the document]
xanaxx: Fixed in latest commit, please check. Test will be done after functions approval.

tests/functional/emit-schema-sdl.ts
```diff
@@ -95,6 +102,26 @@ describe("Emitting schema definition file", () => {
       checkSchemaSDL();
     });
 
+    it("should generate schema SDL file on selected file path", async () => {
+      const targetPath = join(__dirname, "schemas", "graphql", "schema.qgl");
+      await buildSchema({
+        resolvers: [MyResolverClass],
+        emitSchemaFile: targetPath,
+      });
+      expect(pathArg).toEqual(targetPath);
```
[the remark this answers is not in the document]
xanaxx: Updated tests in latest commit. Hope I've made this proper way.

xanaxx:
After all attempts I'll just remove failing tests. There's no clear reason why it is happening. If someone have idea or can do this please feel free to do this. 

More info:
Test is failing if we provide other path than `path.resolve` one, for eg. `string`, `path.join`. On my local machine everything is working and all tests are passing.

xanaxx:
Changed custom functions to `fs-extra` - let's use something well maintained and used.

19majkel94:
@xanaxx 
It's bad that you've decided to go with `fs-extra` - I can't accept that:
https://packagephobia.now.sh/result?p=fs-extra

xanaxx:
@19majkel94
Okay. I'll try to take other attempt on this one. I think I've found problem with last issue which drove me to use `fs-extra`.

xanaxx:
Please check this one. Only using `rimraf` in tests and as `devDependency`.

tests/functional/emit-schema-sdl.ts
```diff
@@ -52,126 +42,235 @@ describe("Emitting schema definition file", () => {
...
-      expect(pathArg).toEqual("testPath1");
-      checkSchemaSDL();
+      const targetPath = path.join(TEST_DIR, "schemas", "test1", "schema.gql");
+      await emitSchemaDefinitionFile(targetPath, schema);
+      expect(fs.existsSync(targetPath)).toEqual(true);
+      checkSchemaSDL(fs.readFileSync(targetPath).toString());
+    });
+
+    it("should call writing file with schema SDL", async () => {
+      const targetPath = "./type-graphql/schemas/schema.gql";
```
19majkel94: Looks like you have a lot of duplicated test - is this a merge error or you write the path by hand on purpose?

19majkel94:
@mszekiel 
I've refactored the filesystem operation helpers to async/await to simplify them and the mocks in tests.

Please check if everything looks right to you and then we can merge it 😉

tests/functional/emit-schema-sdl.ts
mszekiel: It was by design. As I want to be sure if users provide simple string or from `path.resolve`.

MichalLytek: path.resolve returns a string, so there's no difference until the path is correct

mszekiel: I know that but I just wanted to be sure if everything's okay. 🔐  If you think it is not necessary we can remove these tests.

19majkel94: I think that's ok - we should just check if it can write the file in new nested directories and check if we won't swallow an error.

19majkel94:
@mszekiel 

PR #276, opened 2019-03-12
Add types field at package.json, because we do not use default value

galkin:
@19majkel94, I migrated to 0.17 and lost typescripts support at WebStorm. This pull should fix the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 <!-- here goes all the unreleased changes descriptions -->
 ### Features
 - add support for emitting schema file in not existing directory (#269)
+### Fixes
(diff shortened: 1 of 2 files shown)


19majkel94:
@galkin 
I don't have Webstorm but this change doesn't break VSCode support so we can safely merge it 🎉

ForsakenHarmony:
should have been broken in vscode as well 🤔

galkin:
@19majkel94, when will you publish new version?

19majkel94:
@ForsakenHarmony @galkin
It took me a while but `type-graphql@0.17.1` has been released! 🎉

galkin:
@19majkel94, thank you!

Issue #160, opened 2018-10-01
Cannot return plain objects when using inherited ObjectTypes


cuchi:
**Describe the bug**
When I'm using inheritance between `@ObjectType` classes, I cannot return anything that is not an valid `instanceof MyInheritedObjectType`. This does not happen when I use just plain, non-inherited classes.

**To Reproduce**
There is [this gist](https://gist.github.com/cuchi/cbe3f7ab2abb2c5d667f20a154876309)

**Expected behavior**
In the gist I made, `query { users }` plays just nice with the plain objects array I provided. TS also statically checks it with no problem.
When calling `query { inheritedUsers }`, I expect it to have the same behavior, but it instead fails because the array does not meet the correct type.

**Logs**
```
{
  "data": null,
  "errors": [
    {
      "message": "Expected value of type \"InheritedUser\" but got: [object Object].",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "complexUsers",
        0
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "message": "Expected value of type \"InheritedUser\" but got: [object Object].",
          "locations": [
            {
              "line": 2,
              "column": 3
            }
          ],
          "stacktrace": [
            "Expected value of type \"InheritedUser\" but got: [object Object].",
            "",
            "GraphQL request (2:3)",
            "1: {",
            "2:   inheritedUsers {",
            "     ^",
            "3:     id",
            "",
            "    at invalidReturnTypeError (/home/paulo/src/type-graphql-test/node_modules/graphql/execution/execute.js:766:10)",
            "    at completeObjectValue (/home/paulo/src/type-graphql-test/node_modules/graphql/execution/execute.js:758:13)",
            "    at completeValue (/home/paulo/src/type-graphql-test/node_modules/graphql/execution/execute.js:660:12)",
            "    at completeValue (/home/paulo/src/type-graphql-test/node_modules/graphql/execution/execute.js:629:21)",
            "    at completeValueWithLocatedError (/home/paulo/src/type-graphql-test/node_modules/graphql/execution/execute.js:580:21)",
            "    at completeValueCatchingError (/home/paulo/src/type-graphql-test/node_modules/graphql/execution/execute.js:550:12)",
            "    at /home/paulo/src/type-graphql-test/node_modules/graphql/execution/execute.js:684:25",
            "    at Array.forEach (<anonymous>)",
[...]

PR #279, opened 2019-03-13
fix(inheritance): allow for returning plain objects

19majkel94:
Pretty dirty solution but should be enough as a temporary fix until #183 will be achieved 💪 

Closing #160 🔒

### DIFF
--- a/src/schema/schema-generator.ts
+++ b/src/schema/schema-generator.ts
@@ -37,6 +37,7 @@ import {
   GeneratingSchemaError,
   MissingSubscriptionTopicsError,
   ConflictingDefaultValuesError,
+  InterfaceResolveTypeError,
 } from "../errors";
 import { ResolverFilterData, ResolverTopicData } from "../interfaces";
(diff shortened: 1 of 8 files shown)


src/schema/schema-generator.ts
```diff
@@ -209,6 +217,15 @@ export abstract class SchemaGenerator {
...
               return fields;
             },
+            resolveType: instance => {
+              const typeTarget = implementingObjectTypesTargets.find(
+                typeCls => instance instanceof typeCls,
+              );
+              if (!typeTarget) {
+                throw new InterfaceResolveTypeError(interfaceType);
+              }
+              return this.objectTypesInfo.find(type => type.target === typeTarget)!.type;
```
charlex: I've never seen an exclamation mark used in syntax like this.

19majkel94: It's the non-null assertion operator:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html

Issue #327, opened 2019-05-04
Feature: allow returning custom async iterators from `subscribe` method


rfermann:
**Is your feature request related to a problem? Please describe.**
I'm trying to use Prisma subscriptions in conjunction with TypeGraphQL. For that purpose Prisma is providing functions to return async iterators from the `subscribe` method of GraphQL subscriptions. Unfortunately these async iterators cannot be passed through TypeGraphQL because TypeGraphQL currently always creates and returns its own async iterators.

**Describe the solution you'd like**
I would like to have a possibility to create/return custom async iterators within the `Subscription` decorator instead of having TypeGraphQL creating the async iterators.

A solution for that topic could be a new `subscribe` property for the `Subscription` decorator, which takes any async iterators. When set, `generateSubscriptionsFields ` directly returns the content of this property to `fields[handler.schemaName].subscribe` instead of creating its own async iterator.

**Describe alternatives you've considered**
Another solution would be to plug in Prismas PubSub system as described [in the docs](https://typegraphql.ml/docs/subscriptions.html#using-a-custom-pubsub-system). But this alternative has some downsides:
- Prismas PubSub system seems not to be accessible from anywhere except the from the Prisma Client
- I would need to create all the events on my own instead of having Prisma creating events for creating, updating and deleting datasets automatically

**Additional context**
I already have a first draft of this feature working locally. Based on this I can create a pull request for this feature, if you like.

19majkel94:
>  a new subscribe property for the Subscription decorator, which takes any async iterators

It can be a good temporary solution for your problem. But it suffers from lack of access to class properties, so no easy DI possible. 

I am thinking about a new API for subscriptions that is all class based - one `@SubscriptionResolver` class per subscription, field/method for topics or method for `subscribe`, as well as normal resolver method like now. But it may take a while to polish that design and implement it.

> I already have a first draft of this feature working locally. Based on this I can create a pull request for this feature, if you like.

It's not a big change so feel free to create a PR with proper tests added and docs updated about this new feature 😉

19majkel94:
Closed by #328 🔒

PR #328, opened 2019-05-04
Add `subscribe` option to `Subscription` decorator

rfermann:
As described in #327, this PR adds a `subscribe` option to the `Subscription` decorator. This option allows it to use custom subscription systems, e.g. the one Prisma is providing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 <!-- here goes all the unreleased changes descriptions -->
 ## Features
 - add support for creating custom parameter decorators (#329)
+- allow to provide custom `subscribe` function in `@Subscription` decorator (#328)
 
 ## v0.17.3
 ### Features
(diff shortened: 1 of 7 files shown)


rfermann:
> We should have an union in `SubscriptionOptions`

I thought about this for a while now. 2 solutions for that came to my mind:
1. creating a helper type like that:

`type RequireOnlyOne<T, Keys extends keyof T = keyof T> = Pick<T, Exclude<keyof T, Keys>> &
  { [K in Keys]-?: Required<Pick<T, K>> & Partial<Record<Exclude<Keys, K>, undefined>> }[Keys];`

`type SubscriptionOptionsTopicsOrSubscribe = RequireOnlyOne<SubscriptionOptions, "topics" | "subscribe">;`

2. integrate `subscribe` into `topics`.

I'm not that happy with both options, but I haven't found any better solution yet... Do you have a more suitable idea for this?

19majkel94:
@rfermann 
I've added a little type helper from `type-fest` library to achieve this 😉

rfermann:
@19majkel94 thanks for your help. Your solution looks much more professional  than mine 👍

PR #356, opened 2019-06-04
Fix typo depreciation/deprecation

robations:
Tests/typechecks seem to be passing.

Technically someone might refer to the interface externally and could be considered a breaking change, but I would guess this is unlikely?

### DIFF
--- a/src/decorators/types.ts
+++ b/src/decorators/types.ts
@@ -58,7 +58,7 @@ export interface ResolveTypeOptions<TSource = any, TContext = any> {
 }
 export type BasicOptions = DecoratorTypeOptions & DescriptionOptions;
 export type AdvancedOptions = BasicOptions &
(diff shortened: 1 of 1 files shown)


19majkel94:
> Technically someone might refer to the interface externally and could be considered a breaking change, but I would guess this is unlikely?

So export an alias with deprecations in JSDoc and changelog. We will remove it in next major version.

robations:
Cool, will take a look later and update 👍

19majkel94:
@robations 
Here is an example commit with renaming and deprecating:
https://github.com/19majkel94/type-graphql/commit/a3fa121990ee053f21307d414178f8df5b7ae8d8#diff-6c58ae6ccad72d41e6c1b57e29ac7621

19majkel94:
Closing via 38035bd 🔒

PR #363, opened 2019-06-23
docs: getting started update snippet deprecation

audiolion:
findById has been deprecated in TypeORM and the recommendation is to simply use `find(id)` now.


19majkel94:
Thanks for your report!

However, in this snippet I use `recipeService: RecipeService` which is not a TypeORM repository, so it have custom `findAll` and `findById` methods.

So closing this PR 🔒

Issue #228, opened 2019-01-09
Extend a typedef in another file


Xample:
Sometimes you want to add a property to whatever model but you want to keep all the logic (typedef and resolver) in a folder appart. This can be done using the ‘extend’ keyword in a typedef.
https://www.apollographql.com/docs/graphql-tools/generate-schema.html#extend-types   Actually this is more a mixin than an extend. How could you achieve the same using type-graphql ?

19majkel94:
TS mixins are awful and (I hope) a deprecated pattern :confounded:

ES6 class mixins are the only solution:
https://youtu.be/j9dOdjBzARo?t=441

I plan adding to TypeGraphQL a type-safe port of `mix(Foo).with(Bar)` approach:
https://github.com/justinfagnani/mixwith.js#mixwith

Xample:
Okay good so this way you can extend a class in another one, however, unlike an internal module (which would be merged by the language) you still need the parent class to know the subclasses to explicitly extend them.

19majkel94:
> unlike an internal module (which would be merged by the language) 

TS is not able to merge classes in namespaces (internal modules):
```ts
// a.ts
namespace MyApp {
  export class User {
    name!: string;
  }
}

// b.ts
namespace MyApp {
  export class User {
    age!: number;
  }
}
```

![code_2019-01-18_11-04-56](https://user-images.githubusercontent.com/10618781/51380837-ee20d080-1b12-11e9-8e94-d2db8cbd1315.png)

To achieve this, TS would have to support partial classes:
https://github.com/Microsoft/TypeScript/issues/563

Logic-Bits:
Hi all,

TypeORM solved that already with "@ChildEntity()"

https://github.com/typeorm/typeorm/blob/master/docs/entity-inheritance.md

Would be awesome if Type-Graphql could implement it the same since alot is already inspired by the same stuff.
```

import { PrimaryGeneratedColumn, Column, Entity, ChildEntity } from "typeorm";
import { Field, ObjectType } from "type-graphql";
import { InvoiceDocument } from "../documents/invoice.model";

@ChildEntity()
@ObjectType({ description: "" })
export class CustomInvoiceFields extends InvoiceDocument {
    
    @Column()
    @Field({ nullable: true })
    CoolesExtraFeld: String;
}
```

19majkel94:
@Logic-Bits So you propose the same solution as earlier?

```ts
// modules/user/user.ts
@ObjectType()
[...]

Issue #77, opened 2018-05-12
Schema directives support


19majkel94:
## Currently blocked by https://github.com/graphql/graphql-js/issues/1343

---

Ability to define any arbitrary graphql directive to be passed to the compiled schema:

```ts
@ObjectType({ directives: ['@cacheControl(maxAge: 240)'] })
class SampleObject {
  // ...
}
```

Which should be equivalent to:
```graphql
type SampleObject @cacheControl(maxAge: 240) {
  # ...
}
```

It might be also implemented as a separate decorator that would make possible to create custom decorators like `@CacheControl({ maxAge: 240 })`:
```ts
@Directive('@cacheControl(maxAge: 240)')
@ObjectType()
class SampleObject {
  // ...
}
```

jas99:
@19majkel94 Why this is blocked? at face seems less complicated.

19majkel94:
Go to the link in first post and read the discussion :wink	
If you can implement this - feel free to make a PR 😃

ntziolis:
This sadly prevents anybody from being able to use `type-graphql` that is using graphcool/prisma as a backend. We would love to use it but sadly without being able to generate the schema necessary by the downstream tooling its a non starter :(

19majkel94:
@ntziolis 
AFAIK, Prisma is the glue between your GraphQL server and your database.
You define your model using SDL and you receive a generated client that acts as an ORM in your server app.

![](https://i.imgur.com/3VzVxCe.png)

So there should be no problem to use Prisma along with your GraphQL API build with TypeGraphQL. Of course, the tighter integration (defining prisma model and your API GraphQL Object Type in one class with decorators) would help in reducing boilerplate, but I think it's not a deal breaker 😉

19majkel94:
And to report an update about the status of the issue - now I think I understand where is the problem with directives in `graphql-js`. They are purely the SDL thing, so tools written in different languages can parse the schema definition and apply certain behaviors in runtime.

So as with `graphql-js` we define a runtime JS object, other tools (like Prisma or Apollo Engine) don't have an access to the directives. That's why we have to wait for a new GraphQL specification that will expose the directives info via an introspection query. For now the only workaround is to generate `typeDefs` and `resolversMap` instead of executable `GraphQLSchema` and paste the "directives" from metadata storage into proper lines in SDL string 😕

ntziolis:
[...]

PR #369, opened 2019-07-03
Directives support with `@Directive` decorator

j:
I was toying with federation support and going about it the manual way.

Here's the issue I also posted on `apollo-server` https://github.com/apollographql/apollo-server/issues/2965

This doesn't work but if you look at the test, it shows the federated schema.

I sort of am clueless on how to support directives so it'd be cool if someone could chime in on this.

I was mocking the test case from the `federation-demo`'s reviews service.  Ideally we can have a fully functional replica.

/cc @JacksonKearl


JacksonKearl:
Hey I added some stuff to hook together our `buildFederatedSchema` and your directives and it seems to be coming together fairly well: I can view the service in GraphQL Playground and it contains all the required directives/entities/etc. 

The next blocker is that defining
```ts
  @Query(returns => Recipe)
  async __resolveReference(reference: Recipe) {
    return this.items.find(item => item.title === reference.title);
  }
```

causes an error here https://github.com/JacksonKearl/type-graphql/blob/federation/src/schema/schema-generator.ts#L102

j:
@jacksonkearl awesome! I’m on vacation so I’ll look into it later. As for the resolver function, I might make a decorator for that. I also think there’s a way to have the function be whatever name you want and define the graphql name in the options, but I’m not too sure.

j:
Also I believe you need to use ‘@FieldResolver()’ instead of ‘@Query()’. This would go in the resolver class and have to set up the Recipe as the ‘@ObjectType(() => Recipe)’ (I’m typing on a phone and phone coding is hard)

19majkel94:
@j looks great!
But can we make it more elastic and support other directives too? Then the federation directives would be just build on top of them, instead of hardcoded.

Also, the decorators should be prefixed with `Federation`, e.g.: `@FederationExtends()`.

j:
@19majkel94 yup! I actually wanted to do that but decided to just do direct to prototype. I can clean things up when I get back and also do a fully replicated demo hopefully. If I do the directives route, I could change this to @Directive() instead and people can implement federation stuff that way and I’ll just remove the @Federation prefix?

Also, notice how @JacksonKearl did the implementation. Is that okay? What’s the best way to get typedefs and resolvers instead of an executable schema? Also, to remove federation as a dependency, should I make a pluggable printSchema option so people can bring their own? In this case, use the printSchema from the federation library?

I can get these working examples next week. Let me know how you’d like them to be implemented :)

19majkel94:
>  If I do the directives route, I could change this to @directive() instead and people can implement federation stuff that way and I’ll just remove the @federation prefix?

No, we can have a generic `@Directive()` mechanism and then export also a set of custom apollo federation decorators that will call `@Directive` under the hood.

> What’s the best way to get typedefs and resolvers instead of an executable schema?

Current pipeline is metadata -> GraphQLSchema -> derive typeDefs and resolvers. This way we can quite easily support both worlds. We can wrap the require in try-catch, so all that users would have to do is to install the `@apollo/federation` dependency by themselves.

In the future there will be a plugin system that would solves this kind of tricks and complicated setup (own `printSchema` impl).

j:
@19majkel94 great! I’ll work on this. I need to research more on what’s best to use, ‘extensionASTNodes’ or ‘astNode’. ‘extensionASTNodes’ seems like the one that should be used, but would require a chance to federations printSchema function. I haven’t really researched the proper way, since I’m not sure there really is considering there’s no direct support for directives in type constructors.

j:
@19majkel94 It seems that a lot of directive implementations rely on `typeDefs` and `graphql-tools` to create executable schemas with defined directives, however, the route in this PR, federation will only output federation specific stuff, but if someone wants to use other directives, they'll be stripped out because the graphql printSchema strips them away from fields and object types, so unless I'm mistaken, I'm not sure what benefits will come from allowing directives on the schema.

19majkel94:
>  they'll be stripped out because the graphql printSchema strips them away from fields and object types, so unless I'm mistaken, I'm not sure what benefits will come from allowing directives on the schema

As discussed in #77, we have to fork the `printSchema` helper from `graphql-js` to print the directives too.

j:
@19majkel94 Gotcha, in #77, in order to create directives like, 

```javascript
@Directive('@cache(age: "7 days")')
```

You'll need to parse that to create the proper AST, but those aren't exposed I don't think in the parser.  I could parse them manually or just create a "fake" type with directives and just pull the parsed directives out, but it's just being hacky.

The other option is to just do something more like, 

```javascript
@Directive('cache', { age: '7 days' })
```

In the end of the day, I think the most flexible way is if `type-graphql` build out schema as a string and people could bring their own `buildExecutableSchema` from typeDefs and a resolver map.  Do you have ideas on the best way to parse the first directive option, or is the second one okay?

19majkel94:
> The other option is to just do something more like

Yes, if we need to construct AST from the decorator metadata, it should be the API we introduce.

> I think the most flexible way is if type-graphql build out schema as a string typeDef 

This will break all the features that operates on `GraphQLObjectType` and others, like query complexity.

We can't do that until the specification will allow for reading directives metadata and exposing them in `extensions` property of graphql-js runtime objects:
https://github.com/graphql/graphql-js/issues/1343

j:
@19majkel94 I pushed a more generic @Directive using the second option above.  I'm doing this before my flight back home, so there's probably things I missed.  Let me know what you think.

Notes: I moved the auto-magic "directives" initialization of the schema so that users have to pass that part in themselves (to not be hard coded for federation support), this way you can bring in custom directives too.  I think all that's needed then is to build a printer that keeps directives, which I've seen around in various issues.  My previous federation tests are passing.

ldiego08:
@j thank you so much for taking care of this feature! I'm on the edge of my seat seeing how this PR unfolds. Just one thing, I know @19majkel94 requested these decorators to be prefixed with `Federation` but given how Apollo Federation is the collective name of something that's becoming part of the Apollo core, it seems a bit unnecessary and might clutter the decorated classes. In SDL, the Federation directives are named `key`, `extends`, etc., and for me, it makes sense we name them in code-first approach accordingly unless there are naming collision concerns. Any particular reasons why the decorators need to be prefixed?

Here's how this would look like to me in code without the prefix and seems good enough:

```ts
import { ObjectType Key, Extends } from 'typegraphql';

@Extends()
@ObjectType()
class User {
   @Key()
   id: number; 
}
```

Federation directives could go into its own `typegraphql/federation` folder if there's a definitive need to have them apart. I know there are plans for adding support for TypeORM, Prisma, and other packages, so it might be a good way forward or else decorators for all those will need prefixing.

19majkel94:
> Any particular reasons why the decorators need to be prefixed?

Because to support #228 we might introduce the `@Extends(base => User)` decorator that will

>  or else decorators for all those will need prefixing

That's correct, there will be `@PrimsaModel()`, etc.

JacksonKearl:
In terms of conflicts, it would be possible to `import as`:

```ts
import { ObjectType, Key, Extends } from 'typegraphql';
import { Key as FederationKey, Extends as FederationExtends } from 'typegrapql/federation'
import { Model as PrismaModel } from 'typegraphql/prisma'

@Extends()
@ObjectType()
@FederationExtends()
@FederationKey({fields: "id"})
class User {
   @Key()
   id: number; 
}
```

But up to you if that's something you'd want.

j:
@JacksonKearl @ldiego08 You can also do the inverse of the "as" . It's sort of why I namespaced it as 

```typescript
// decorators/index.ts
export { Federation } from './federation';

// my resolver
import { Federation } from './federation';

const { Key, Extends, ... } = Federation;
```

But I'm not opposed to prefixing with Federation*.  Another note, since they are simply directives, the `Federation*` exports aren't really even needed since you can just do:

```typescript
@ObjectType()
@Directive('key', { fields: "id" })
class UserResolver {
  // ...
}
```

Also, this isn't supported in this PR:
```typescript
class User {
   @FederationKey()
   id: number; 
}
```

I'll try to touch more things up today.

j:
I'm still working on this, but decided to see if the apollo team would take a PR to support passing in already constructed schemas to make this implementation a lot cleaner.

https://github.com/apollographql/apollo-server/pull/3013

This would make federation support really simple and not have to even be a dependency.  I believe for other directives, you can use `attachDirectiveResolvers` from `graphql-tools` and don't really need to add a custom printer.  I'll prototype this too.

/cc @JacksonKearl @19majkel94 @ldiego08

j:
Update: I’m able to replicate the entire federation demo using this PR with my PR for @apollo/federation.

I’ll push the examples after cleaning things up.  One thing that feels awkward is defining externals on fields in a federated schema. I’ve been so used to those properties in the class to need to be filled, but noting them as readonly with undefined types makes me feel better. Another weird thing was the reference resolvers, the Apollo code samples on their demo use the world “object” for the root on all field resolvers which makes me feel that they are the same thing, but root is actually the type name with federation key. I’ll make it clear in the examples.

j:
I pushed examples using the PR to `@apollo/federation` (apollographql/apollo-server#3013).  Check it out.  CI failed due to the upstream `@apollo/federation` change.  If that can get merged, then federation support is ready to go, otherwise we'll have to go the route of converting to `SDL` -> schema -> `SDL` -> schema (which feels nasty).  I feel like this is as elegant as it can get until directives are baked in which should be in the next major `graphql-js` version.  They may add a `GraphQLSDLDirective` type for printSchema.

https://github.com/19majkel94/type-graphql/pull/369/commits/9bbed68c47cc31015cb9f06bb0d2b68047db7db7

Next on the list is that I'm going to play around with 3rd party directives and see if I can get those to work natively with this change.

/cc @JacksonKearl @19majkel94 @ldiego08

j:
PS, not gonna lie, this feels reaaaaaal good.  :fire:

The only "gotcha" I found is that `__resolveReference` isn't a true GraphQL field resolver.  So when you use `@Root()`, it's not the actual object type, but instead is representation reference sent by the gateway ([@apollo/federation/src/types.ts#L98](https://github.com/apollographql/apollo-server/blob/master/packages/apollo-federation/src/types.ts#L98)).  I was using `@Root() reference: Partial<MyType>` instead of actually mapping it to the class since it's not really the class.  I'm not sure how I feel about this, but the examples work.

j:
I came into an issue regarding my previous comment.  You can't get `@Ctx()` into `__resolveReference`.  It looks like `@Ctx()` is the info object.

j:
@19majkel94 I made a change to use directive parsing, as well as more complete generic directive tests, etc.  I'd like your opinion on a possible solution regarding `__resolveReference`.

The way `graphql-tools` creates executable schemas, they check if fields start with `__` and if they do, don't add it as a GraphQL field, but instead set it as a `GraphQL*Type` field.  `__` is reserved anyway so it shouldn't break anything.

I have a few ideas and would like to hear what you think.

1. If fields start with `__`, automatically remove it as a field resolver and make it a field on the constructed type (this is what `apollo-tools` does.
2. Create a new `@ObjectTypeField` decorator.

The reason one of the above is needed, is if we want `__resolveReference` to work for federation and `@Resolver` classes, then we have to support it using one of the above methods.  Why it doesn't work at the moment (well it does, until you need `@Context`) is because `@apollo/federation` and `graphql-tools` takes resolvers prefixed with `__` and turns them into an object type field, from there, `@apollo/federation` creates other resolvers to reference this field and their arguments differ from what `type-graphql` does.

Let me know your thoughts!

j:
@19majkel94 ping

19majkel94:
I think we can go with the common behavior that `__` prefix means it's a object type config property, not a field resolver.

And do we need to make federation a dependency? Maybe we can just point in docs and examples that users should manually print the schema using the federation print version, rather than using standard apollo buildTypeDefsAndResolvers?

Also, why you have give up on the `@Directive('cache', { age: '7 days' })` syntax? It's much better than the string based one.

j:
@19majkel94 thanks for the response!  Our backend is feature-locked until federation is supported.  I can work on anything as quickly as possible so we can keep moving forward on our backend :)

I can switch back to that type of directive instead of string based.  The reason I ended up doing it is because I have to call a `parse*` method anyway.  After I wrote tests for the directives and used a 3rd party directive library, they were looking at the `ValueNode` for the directive definition and I originally had everything working with hard-coded `StringValue`.  However, I can just use `parseValue` on the `@Directive` arguments to get it.  I'll switch back tomorrow.

Also, it looks like `extensions` is going to be implemented in `graphql-js` soon, so when that does, I'll send PR requests to federation to switch execute `__resolveReference` from the extensions prop, and later when Directives are baked in, I can rewrite all the necessary code.

Before I move all `@Field()` and `@FieldResolver()` fields prefixed with `__` to the object type, how do you propose I get instances of `@FieldResolver()` owning class from DI?  `container.getInstance` requires `ResolverData` to be present.  Should I make that optional?  Since it's going from being a resolver to not being a resolver, that is.  I still want to be able to make use of DI on resolver classes.

j:
@19majkel94 

> And do we need to make federation a dependency? Maybe we can just point in docs and examples that users should manually print the schema using the federation print version, rather than using standard apollo buildTypeDefsAndResolvers?

No. The way I ended up doing it (when my PR gets accepted in `@apollo/federation`) makes it so that federation doesn't need to be a dependency.  I can just write in the docs on how to create a federated schema.  You don't even need to use their printSchema or anything.  My PR allows for ANY GraphQLSchema to become federated... pretty cool!  That's why these tests are failing.  I'll probably remove the federation test since the route I chose is just generic directive support.

19majkel94:
> Our backend is feature-locked until federation is supported. I can work on anything as quickly as possible so we can keep moving forward on our backend :)

Feel free to use your fork and wait for a proper support in TypeGraphQL. I don't want to rush and introduce an incomplete feature and the propose a breaking changes that "fixes" it.

Ast nodes work is great but with extensions it would be more elegant solution:
https://github.com/graphql/graphql-js/issues/1527#issuecomment-511991379

> I'll switch back tomorrow.

Can we support both approaches? Or is it hard to distinguish them?
With args it should be a problem:
- `@Directive('@cache(age: "7 days")')`
- `@Directive('cache', { age: '7 days' })`

But without args it might be complicated:
- `@Directive('@authorized')`
- `@Directive('authorized')`

It's just a matter of stripping/adding the `@` prefix, so it would be nice to support them both 😉 

> how do you propose I get instances of @FieldResolver() owning class from DI?

The same as now for other resolvers.

> Since it's going from being a resolver to not being a resolver, that is. I still want to be able to make use of DI on resolver classes.

What do you mean? It's called only once before building a schema? Or just like normal resolver but it doesn't have access to `args` 2nd parameter?

j:
> Technically a 3rd party library could require a resolver called, `__collection` that expects args to be `(client: MongoClient)`  which expects no arguments (example).  Are you thinking that `__*` resolvers are always resolvers without `args` ?  If that's the expectation, I can certainly do that, but I believe the argument signature can be anything.  Which is why I was wondering how to get class instances from DI when/if they don't have the typical resolver signature.  I was going to just have `__` be plain functions with no decorator support.


> Can we support both approaches?

Yeah I can allow both approaches.  Shouldn't be too difficult.

src/utils/printSchema.ts
```diff
@@ -0,0 +1 @@
+export const defaultPrintSchemaOptions: PrintSchemaOptions = { commentDescriptions: false };
```
Superd22: `PrintSchemaOptions` isn't imported and causes the transpilation to fail, just fyi

j: Yup, forgot to remove this file from this PR.  I'll rebase everything later.  Removed in the last commit.

j:
@19majkel94 I realized I rebased and lost everything up until the revert. I rewrote tests that hopefully are a little more clear and full.  Latest updates allow SDL along with non-SDL.

Next up is moving field resolvers & fields prefixed with `__` to the object type.

Let me know what you think of the latest changes.

19majkel94:
>  I believe the argument signature can be anything.

Right, I was thinking only about the `__resolveReference` case. So we should detect usage of the decorators in `__` fields resolvers and warn about that.

So as we might don't have an access to the `ctx`, we have no access to the scoped container. So without an instance we have no safe `this` access and no decorators, so maybe placing this kind of resolvers in the `@ObjectType` decorator is a better idea? It will be easy to make it typesafe by interface augumentation, even in a plugin system.

```ts
@Directive('key', { fields: "id" })
@ObjectType({
  resolveReference(reference: UserReference, { fetchUser }, info): Promise<User> {
    // ...
  },
})
class User {
 // ...
}
```

j:
@19majkel94 Yeah, it's a matter of supporting the "apollo" world use-cases or just ignoring them but having documentation on how to add those features.

Just having this PR as it stands gives apollo support (if my PR gets merged).

It's not hard after you create a `type-graphql` schema to just do this:

```typescript
import { addResolversToSchema } from 'apollo-graphql';

  resolvers: [RecipeResolver]
});

// Add Federation Magic Here.
addResolversToSchema(schema, {
  Recipe: {
    async __resolveReference(object, { fetchRecipe }): Promise<Recipe> {
      // ...
    }
  }
});
```

Or can create a new argument in `buildSchema` like `resolverMap` or `plainResolvers` if you want this to be a shortcut in the schema building process as well.

j:
@19majkel94 My vote is to leave as is and people can use `addResolversToSchema` for now, and we can work on `extensions` later when it's more adopted.

With that said, is there any more you think I should do to this PR (besides remove federation examples until my PR is merged?)

I can add "directive" documentation if you feel that it's needed or we can wait until it's possible to attach federated types to an existing schema for me to finish everything up.

j:
@19majkel94 they should get to my PR next week so by end of week we can have federation fully functional. :)

ldiego08:
@j it'd be amazing to add documentation if possible. I'll be using this first second this is out and would be nice to have some examples. 😄

j:
@ldiego08 I can do that.  The tests pretty much tell the whole story as well.  But if you're talking about federation integration when this comes out, that'll be different.  I can add documentation for both, though. :)

ldiego08:
> @ldiego08 I can do that. The tests pretty much tell the whole story as well. But if you're talking about federation integration when this comes out, that'll be different. I can add documentation for both, though. :)

Oh, I just realized this PR only introduces custom directives, but not federation-specific decorators. I get what you're saying on documentation now. :)

19majkel94:
> My vote is to leave as is and people can use addResolversToSchema for now, and we can work on extensions later when it's more adopted.

Yes, as this PR is only about directives support, we can leave it now and wait for the `extensions` support with a proper apollo federation plugin without `addResolversToSchema` workarounds.

src/schema/schema-generator.ts
```diff
@@ -255,12 +268,17 @@ export abstract class SchemaGenerator {
...
       return {
         target: objectType.target,
         isAbstract: objectType.isAbstract || false,
         type: new GraphQLObjectType({
           name: objectType.name,
           description: objectType.description,
+          astNode: this.createClassDirectiveNodes(
+            objectType.name,
+            objectType.directives,
+          ) as ObjectTypeDefinitionNode,
```
19majkel94: Why we have to make `as` assertions for every astNode?

src/schema/schema-generator.ts
```diff
@@ -604,4 +638,104 @@ export abstract class SchemaGenerator {
...
+
+      if (definitions && definitions.length > 0) {
+        definitions.forEach(def => {
+          if (def.directives && def.directives.length > 0) {
+            directives = [...directives, ...def.directives];
+          }
+        });
+      }
+    } catch (err) {
+      /** noop */
```
19majkel94: why noop? we should throw error so the user can fix his directive sdl

src/schema/schema-generator.ts
```diff
@@ -604,4 +638,104 @@ export abstract class SchemaGenerator {
...
+
+    if (!nameOrSDL.startsWith("@")) {
+      return [
+        {
+          kind: "Directive",
+          name: {
+            kind: "Name",
+            value: nameOrSDL,
+          },
+          arguments: Object.keys(args).map(arg => ({
```
19majkel94: `argKey`

src/utils/emitSchemaDefinitionFile.ts
```diff
@@ -10,7 +10,6 @@ const generatedSchemaWarning = /* graphql */ `\
 # !!! THIS FILE WAS GENERATED BY TYPE-GRAPHQL !!!
 # !!!   DO NOT MODIFY THIS FILE BY YOURSELF   !!!
 # -----------------------------------------------
-
```
19majkel94: Why you've removed the blank line separation?

package.json
```diff
@@ -18,8 +18,8 @@
     "postinstall": "node ./dist/postinstall || exit 0"
   },
   "peerDependencies": {
-    "@types/graphql": "^14.0.7",
-    "graphql": "^14.1.1"
+    "@types/graphql": "^14.4.1",
+    "graphql": "^14.4.1"
```
19majkel94: We should place the breaking change warning in changelog

src/decorators/Directive.ts
```diff
@@ -0,0 +1,41 @@
+import { MethodAndPropDecorator } from "./types";
+import { SymbolKeysNotSupportedError } from "../errors";
+import { getMetadataStorage } from "../metadata/getMetadataStorage";
+import { DirectiveMetadata } from "../metadata/definitions/directive-metadata";
+
+export interface DirectiveArgs {
+  [arg: string]: any;
```
19majkel94: `Record<string, any` is enough

src/decorators/Directive.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export function Directive(sdl: string): MethodAndPropDecorator & ClassDecorator;
+export function Directive(
+  name: DirectiveMetadata["nameOrSDL"],
+  args?: DirectiveMetadata["args"],
```
19majkel94: Can we avoid the indirect types addressing and use type value (`Record<string, any>`) or named types (`DirectiveArgs`). It doesn't match the current code style and I see it everywhere in your PR.

src/decorators/Directive.ts
```diff
@@ -0,0 +1,41 @@
...
+  return (targetOrPrototype, propertyKey, descriptor) => {
+    const directive = { nameOrSDL, args: args || {} };
+
+    if (!propertyKey) {
+      getMetadataStorage().collectDirectiveClassMetadata({
+        target: targetOrPrototype as Function,
+        directive,
+      });
+
+      return;
```
19majkel94: use `else` instead of empty early return

src/metadata/definitions/directive-metadata.ts
```diff
@@ -0,0 +1,14 @@
+export interface DirectiveMetadata {
+  nameOrSDL: string;
```
19majkel94: We need a better name than `sdl`... `directiveDefString`? `directiveDefinition`? `directiveTypeDef`? Ideas? 😄

ldiego08: `nameOrDefinition`?

src/decorators/ObjectType.ts
```diff
@@ -1,6 +1,7 @@
 import { getMetadataStorage } from "../metadata/getMetadataStorage";
 import { getNameDecoratorParams } from "../helpers/decorators";
 import { DescriptionOptions, AbstractClassOptions } from "./types";
+import { ObjectClassMetadata } from "../metadata/definitions/object-class-metdata";
```
19majkel94: Looks like not used here

tests/functional/directives.ts
```diff
@@ -0,0 +1,655 @@
...
+          return result.toUpperCase();
+        }
+
+        return result;
+      },
+    });
+  }
+}
+
+class UpperCaseDirective extends SchemaDirectiveVisitor {
```
19majkel94: Do we really need a custom directive in tests? Can't we use something from npm that works the same?
https://www.npmjs.com/package/@saeris/graphql-directives

tests/functional/directives.ts
```diff
@@ -0,0 +1,655 @@
...
+
+        @Query()
+        @Directive("@append @upper")
+        queryWithUpperAndAppendSDL(): string {
+          return "hello";
+        }
+
+        @Mutation()
+        @Directive("foo")
+        mutationWithDirective(): string {
```
19majkel94: mutation and queries are handled 1:1 the same, there's no need for additional tests for them

tests/functional/directives.ts
```diff
@@ -0,0 +1,655 @@
...
+      it("should add directives to query types with arguments", async () => {
+        const queryWithDirectiveWithArgs = schema.getQueryType()!.getFields()
+          .queryWithDirectiveWithArgs;
+
+        assertValidFieldDirective(queryWithDirectiveWithArgs.astNode, "bar", { baz: "true" });
+      });
+
+      it("calls directive 'upper'", async () => {
+        const query = `query {
+        queryWithUpper
```
19majkel94: formatting issues all the way down

tests/functional/directives.ts
```diff
@@ -0,0 +1,655 @@
...
+          withInput: "hello, WORLD!",
+          withInputUpper: "HELLO, WORLD!",
+          withInputOnClass: "hello, WORLD!",
+          withInputUpperOnClass: "HELLO, WORLD!",
+        });
+      });
+    });
+  });
+
+  describe("@cacheControl", () => {
```
19majkel94: Maybe it should be in separate test file? I'm not a fan of all-in-one test suites

tests/functional/directives.ts
```diff
@@ -0,0 +1,655 @@
...
+        ) as GraphQLInputObjectType).getFields();
+
+        expect(fields).toHaveProperty("append");
+        expect(fields.append).toHaveProperty("astNode");
+        assertValidFieldDirective(fields.append.astNode, "upper");
+      });
+    });
+
+    describe("ObjectType", () => {
+      it("calls object type directives", async () => {
```
19majkel94: do we have tests for queries?

tests/functional/directives.ts
```diff
@@ -0,0 +1,655 @@
...
+        expect(data).toHaveProperty("objectType");
+        expect(data.objectType).toEqual({
+          withDirective: "withDirective",
+          withDirectiveWithArgs: "withDirectiveWithArgs",
+          withUpper: "WITHUPPER",
+          withUpperSDL: "WITHUPPERSDL",
+          withAppend: "hello, world!",
+          withAppendSDL: "hello, world!",
+          withUpperAndAppend: "HELLO, WORLD!",
+          withUpperAndAppendSDL: "HELLO, WORLD!",
```
19majkel94: It's handy, so if there's no append and upper 3rd party directives, let's move the directives visitors code to test helpers file.

tests/functional/directives.ts
```diff
@@ -0,0 +1,655 @@
...
+  buildSchema,
+  ObjectType,
+  Mutation,
+} from "../../src";
+import { getMetadataStorage } from "../../src/metadata/getMetadataStorage";
+import Maybe from "graphql/tsutils/Maybe";
+import { SchemaDirectiveVisitor } from "graphql-tools";
+import { ApolloServer } from "apollo-server";
+
+class UpperCaseType extends GraphQLScalarType {
```
19majkel94: I though it's recommended to use create custom scalars this way:
https://github.com/19majkel94/type-graphql/blob/0103091f903f9a38e7d94aed1be9ecf51923b6ca/src/scalars/isodate.ts#L3-L4

tests/functional/directives.ts
```diff
@@ -0,0 +1,655 @@
...
+  visitInputFieldDefinition(field: GraphQLInputField) {
+    if (field.type instanceof UpperCaseType) {
+      /* noop */
+    } else if (
+      field.type instanceof GraphQLNonNull &&
+      field.type.ofType instanceof GraphQLScalarType
+    ) {
+      field.type = new GraphQLNonNull(new UpperCaseType(field.type.ofType));
+    } else if (field.type instanceof GraphQLScalarType) {
+      field.type = new UpperCaseType(field.type);
```
19majkel94: Oh, looks like it's not a scalar but a custom transform trick 😕

examples/apollo-federation/accounts/index.ts
```diff
@@ -0,0 +1,87 @@
+import {
```
19majkel94: I think we should remove the federations example from this PR. We can use it later in separate PR when adding federation examples when the extensions stuff will be resolved. For now it should be usable with directives and addResolversToSchema workaround.

j:
@19majkel94 I'll get on with all the changes soon and remove federation specific stuff.

src/schema/schema-generator.ts
j: Removing it gave me a type error due to the directives needing to be ObjectTypeDefinitionNode or InputObjectTypeDefinitionNode.

src/schema/schema-generator.ts
j: I'm throwing an error below if there are no directives, but yes, I can create a specific parsing error.

src/metadata/definitions/directive-metadata.ts
j: `nameOrDefinition` it is.  Or I can be explicit and have `name?: string;` and `definition?: string`

ldiego08: I’d go with `nameOrDefinition` since it’s not intuitive enough which one gets excluded if you go with two fields. Does `name` overrides `definition` or the other way around? Will there be a scenario where you have both?

ldiego08: I’d go with `nameOrDefinition` since it’s not intuitive enough which one gets excluded if you go with two fields. Does `name` overrides `definition` or the other way around? Will there be a scenario where you have both?

src/schema/schema-generator.ts
ldiego08: Would it be too hard to change the return type of the `createFieldDirectiveNodes` function? It can be piped if backwards compatibility is a concern.

tests/functional/directives.ts
ldiego08: It’d be amazing if this repo integrated prettier to avoid formatting issues. Would you be open to this? @19majkel94 I could work on it.

src/metadata/definitions/directive-metadata.ts
19majkel94: You would have to do the parse phase in decorators body, which is not a great separation tho.

src/schema/schema-generator.ts
19majkel94: I think we can remove some assertions by using proper overloads. Maybe a generic would looks better than assertion where it has to be?

tests/functional/directives.ts
19majkel94: Prettier is integrated with this repo and a commit hook. But prettier doesn't format string, it only recognise `gql` template tag.

j: VSCode copy/paste kept doing this.

tests/functional/directives.ts
j: I initially was building the metadata in it's own build call after everything else was built, so I had to make sure queries and mutations were caught.  But I since then moved it back to building within each type's metadata build steps.  I can remove mutation tests, but if it for some reason gets moved then things'll break.

tests/functional/directives.ts
j: Yeah, I initially was using 3rd party directives but removed to keep it simple, but I can refactor tests to use 3rd party.

19majkel94: We should use 3rd party if exists and if not, just move the them to the helpers file to not make this test suite 1000+ lines big.

j: Also, I did what the docs do for creating directives: https://www.apollographql.com/docs/graphql-tools/schema-directives/

And I was going to use that library you noted but they don't support input types.  :-\

tests/functional/directives.ts
j: I got most of my code from the `apollo` directives section on their website.

tests/functional/directives.ts
j: Yup, all above.  Queries, Mutations, ObjectTypes and InputTypes atm.

tests/functional/directives.ts
j: @19majkel94 yeah, again, it's how the `apollo` documentation does it.

j: @19majkel94 let me know if you think there's another way of doing so.  I went around to a few repositories and others do it like this as well.

tests/functional/directives.ts
j: @19majkel94 done

tests/functional/directives.ts
19majkel94: Ok, so can we just move it to a separate file?

tests/functional/directives.ts
j: @19majkel94 done

j:
@19majkel94 cleaned everything up and made appropriate changes.  Let me know if something may have been missed.

/cc @Superd22 @ldiego08

tests/functional/directives.ts
19majkel94: I don't tests many things that works for mutation too because they are 1:1 the same as query, so it's not a problem to remove the duplicated tests. They are a maintenance problem as we have to fix tests twice because of that.

tests/functional/directives.ts
19majkel94: I have not enough experience with directives as I don't use them. It's ok for me if it's a standard thing.

src/schema/schema-generator.ts
j: @19majkel94 @ldiego08 I ended up creating a new method for each type and removed the whole reverse() then concatenating thing.  So `@Directive()` only supports one directive per annotation which I think is cleaner anyway.

tests/functional/directives.ts
j: I already did in the latest push.

j:
There's a couple use-cases that aren't supported, but I'm afraid to add support for some of them.

- Extending Types
- Interfaces
- Enums
- Unions
- Subscriptions
- Fragments
- Variable Definitions

It seems like a big list, but it shouldn't be hard to add partial support to it.  I'm not entirely sure what happens when a type implements an interface or a type extends another object type.

```graphql
interface Base @something {
  id: ID!
  name: String!
}

type Book implements Base {
  id: ID!
  name: String!
}

type Movie implements Base {
  id: ID!
  name: String!
  reviews: [Review]
}
```

In the above example, I'm not sure if `@something` only applies to the interface fields or if it applies it to every field in the type that's implementing it.  I don't really think many people are using directives in this way anyway (full modification of all fields).  Most people that apply directives to interfaces or on the type level are just adding metadata using SDL or adding a field
to the type.

Also, same for this example:
```typescript
@ObjectType()
@Directive("@lang")
class Base {
    @Field()
    @Directive("@format")
    createdAt: Date
}

@ObjectType()
class Book extends Base {
  // ...
}
```

Does `@lang` apply to all fields of `Book` or just the fields of `Base`?

None-the-less, 100% of federation should be supported with this implementation.  And almost any example I've seen regarding directives as well.  I could, however, patch in extending other object types, I'd just like to hear how others would want this to work?

1. Apply class-level `@Directive` of extended `*Type` to entire `*Type`?
2. Apply class-level `@Directive` of extended `*Type` to the base type?

Or just ignore for now and see what the community requests.  I can at least allow base classes to have `@Directive`s on fields and ignore class-level on base object types.

And since you're going through a re-write soon, the less I add, the better probably (unless there's actual functionality that's blocking), it might be best to keep things to a minimum and I can help with `vNext` later.

19majkel94:
> Subscriptions

These are the same kind of GraphQL type as queries and mutations - `type Subscription` object type.

>  I'm not entirely sure what happens when a type implements an interface

We are just placing the directive syntax on the `interface` definition - what happens when a type implements an interface is a matter of the directive implementation, not our job.

>  I'm not sure if @something only applies to the interface fields or if it applies it to every field in the type that's implementing it.

Again, not our job - it's a decision of the directive implementator.

And in case of extending - field directives are reapplied on inherited field in child class. Class directives are reapplied on the child class.

ldiego08:
> > Subscriptions
> 
> These are the same kind of GraphQL type as queries and mutations - `type Subscription` object type.
> 
> > I'm not entirely sure what happens when a type implements an interface
> 
> We are just placing the directive syntax on the `interface` definition - what happens when a type implements an interface is a matter of the directive implementation, not our job.
> 
> > I'm not sure if @something only applies to the interface fields or if it applies it to every field in the type that's implementing it.
> 
> Again, not our job - it's a decision of the directive implementator.
> 
> And in case of extending - field directives are reapplied on inherited field in child class. Class directives are reapplied on the child class.
> 
> But I would not support extending/inheritance for now as the codebase is not designed for that and I'm refactoring it. For now we only need subscriptions and interfaces support.

Right. Most of these things are out of the scope of TypeGraphQL. This library's only concern is to produce a valid schema from decorators, but from there it's all vanilla GraphQL.

j:
Yup, that was already my assumption from the get-go.  I just was hitting a point where I'll have to continue to research what all directives can be applied to and add appropriate tests.  It's not the most simple thing to do :P

jshearer:
Out of curiosity, are there any updates on this? We're really interested in first-class schema federation support

j:
@jshearer still need the Apollo team to merge my PR. Give that issue a thumbs up and comment :)

snys98:
@j , 👍 could you provide your PR link? I'm going to thumb it up!

j:
@snys98 https://github.com/apollographql/apollo-server/pull/3013

j:
Update: extensions was finally released in graphql, so maybe wait for what the community ends up doing with decorators or move forward with this...

ldiego08:
@j I’m curious how does that impact this PR.

MichalLytek:
We have some movement on apollo side:
https://github.com/apollographql/apollo-server/pull/3013#issuecomment-528494083

But I think we can try to merge the directives support for now with some guides in comment how to manually apply federation. Then we can wait for the new apollo federation and provide a better integration with proper docs :wink:

pierreis:
@j Is there anything we can do to help this pull request move forward?

voslartomas:
@j Hello Jordan, are you using federation already with typegraphql? Are you going to finish this PR, can't wait to use it and this approach seems finished already...

j:
@voslartomas I pushed some of the requested changes.

@MichalLytek regarding the AST name problem, how do you propose I get the actual type name.  All the AST was sort of "filler" since it's all kind of a hack anyway.

Regarding time on this, it's the end of the year and we do e-commerce, and I've been super busy, so I can't really devote too much time to federation at the moment.  But it's not hard to pick up where I left off and add other / change things.

bbenoist:
@MichalLytek I think that the PR https://github.com/j/type-graphql/pull/1 I've just opened includes all the changes you've requested. Please let me know if there is something wrong so that we can finally have this feature in type-graphql 🚀

j:
@bbenoist thanks! looks great /cc @MichalLytek

christopher-avila:
Wow guys. I just arrived to this thread looking for any way to get federation implemented using type-graphql and took me like 3 hours to try to get everything you've discussed or said. It's really impressive job @j . Thank you so much for your time! 

Anything can be done to accelerate the merges on apollo-server any other needed?

Thank you again for your time and effort!

voslartomas:
@MichalLytek Seems ok? Let's merge it, can't wait to use it. Stitching is just wrong, so this feature is crucial.

MichalLytek:
> @MichalLytek Seems ok?

As you can see:

image

No, it's not ok, even the docs are incomplete.

I will try to tackle this PR this week 😉

MichalLytek:
@GeoffreyHervet @Strato @pgueriot @ppinel @christopher-avila @dachristo 
Instead of blindly spamming with aprovals, please check the added part of the docs in terms of language and clarity of message 😉

christopher-avila:
Is not blindly if the goal is to speed up this feature. Since the changes were cosmetic and docs related, I see that "blindly" is a really dared adjective :)

By the way, great job and really thank you!

> @GeoffreyHervet @Strato @pgueriot @ppinel @christopher-avila @dachristo
> Instead of blindly spamming with aprovals, please check the added part of the docs in terms of language and clarity of message 😉

j:
@MichalLytek those docs are 🔥 

@christopher-avila @GeoffreyHervet @Strato @pgueriot @ppinel @christopher-avila @dachristo You should try working on top of this PR to verify it suits everyone's needs.

docs/directives.md
```diff
@@ -2,35 +2,49 @@
...
+```
+
+- object based - using a JS object to pass the named arguments
 
 ```typescript
-import { Directive, Field, ObjectType } from "type-graphql";
+@Directive("deprecated", { reason: "Use newField" }) // syntax without `@` !!!
+```
 
+The the `@Directive` decorator can be placed over the class property or over the class itself, depending on the needs and the placements supported by the implementation:
```
bbenoist: @MichalLytek There is a minor typo here 😉

bbenoist:
@MichalLytek 

> @christopher-avila @GeoffreyHervet @Strato @pgueriot @ppinel @christopher-avila @dachristo You should try working on top of this PR to verify it suits everyone's needs.

On my side, it's working fine. I use it to tag elements with the accounts-js `@auth` directive ❤️

docs/directives.md
MichalLytek: Spend 1 minute looking for the typo before finding "The the" 😄

christopher-avila:
I'll try this PR with Federation, but it really looks awesome. Great, great job with this feature.

Thank you so much!

Superd22:
We've actually been using this PR with a couple tweaks (custom @ResolveReference decorator) on our new project for a few weeks now, it's' working great with Federation.

voslartomas:
@Superd22 How does it work without buildFederatedSchema, are you using only buildSchema from type-graphql, or is there any other step?

Superd22:
> @Superd22 How does it work without buildFederatedSchema, are you using only buildSchema from type-graphql, or is there any other step?

We're using both this branch of typegraphql and the buildFederatedSchema's branch of apollo server together, to great effect :)

MichalLytek:
I've rebased the branch on master, fixed a few things and refactored a bit of code. Now it looks ok and I think that we're ready to go 🎉

christopher-avila:
Great news!!! Congrats!

MichalLytek:
Thank you very much @j for the work and @bbenoist for the help! ❤️

pierreis:
Looks awesome, thanks a lot for the outstanding work! So if I understand correctly, I can only use this for federation with the buildFederatedSchema branch of apollo-server?

MichalLytek:
Published as `v0.18.0-beta.3` - `npm i type-graphql@beta`

MichalLytek:
> I can only use this for federation with the buildFederatedSchema branch

@pierreis 
Technically you can use `createResolversMap` from `type-graphql` and `printSchema` from `@apollo/federation` to use `buildFederatedSchema` with `{ typeDefs, resolvers }`.

I've recreated the apollo federation example from this PR using this trick but despite the fact that `printSchema` print the schema with federation directive, the gateway doesn't see the additional fields. @j can you look at the branch `apollo-federation`?
https://github.com/MichalLytek/type-graphql/tree/apollo-federation

MichalLytek:
Looks like the implementation is buggy - `@Directive(`@key(fields: "upc")`)` works but `@Directive(`@key(fields: "upc")`)` not - `parseValue` doesn't handle properly different types of args like boolean, string, null, etc.

Also we have no support for directives on field resolvers, so the example doesn't want to work.

That's why I don't like rush and "push push push to production we need that feature". Thanks god I've published it as a beta, not as stable release...

MichalLytek:
> It's been 4 months, this PR was opened, I would not call it rush...

This PR initially was about apollo federation and we still doesn't have a proper support from apollo-server. So we were iterating to remove federation hacks and introduce only directives support that could be use for custom federation integration.

MichalLytek:
@pierreis @voslartomas @christopher-avila 
I've fixed the field resolvers and added an example of apollo-federation (a modified federation demo from @j PR):
https://github.com/MichalLytek/type-graphql/tree/20c81f4f7ee82779595002d25784fce3a8ff8b9b/examples/apollo-federation

christopher-avila:
As far as I'm testing it, it works like a charm! I'm migrating our microservices on the typescript stack to Type-graphql, but the first integration is working perfectly. Thank you so much @MichalLytek @j and other people who worked on this.

mkuczera:
Hi, i´m right now confused. Does the apollo-federation implementation also enable the usage of custom directives like @intl?
https://www.apollographql.com/docs/graphql-tools/schema-directives/

Sorry for disturbing, just subscribed to the PR because of the Directive support and ran into confusion regarding apollo-federation

MichalLytek:
Not with current workaround - we use their `printSchema` util that strips off other, non-federation directives. 

Please ping on https://github.com/apollographql/apollo-server/pull/3013 - we need this to have a real federation support without workarounds that breaks features like directives or query-complexity.

fullofcaffeine:
Hi!

Just found out about GraphQL directives and a Google search led me to this issue. Great work on implementing it! I'm a bit confused on the benefits of using it for, say, user authentication + authorization vs the regular way through middlewares. 

Also, the docs here: https://typegraphql.ml/docs/next/directives.html are helpful but the last part showing the runtime implementation seems incomplete. It'd be great to have an example of authorization at least, on how to check the actual data for the directive and doing something based off it.

Thanks!

MichalLytek:
@fullofcaffeine 
> I'm a bit confused on the benefits of using it for, say, user authentication + authorization vs the regular way through middlewares.

If you implement the logic, you should just use middlewares 😉 

Directives are for 3rd party integration, like Apollo Federation or Apollo Cache Control. So if you used some [auth directives](https://github.com/graphql-community/graphql-directive-auth) and while migrating to TypeGraphQL you don't want to reimplement that logic, you can just place the directive decorator and use the same approach as earlier.

fmquaglia:
Hello, thanks a lot for implementing this. <3
Is there an estimated date to release this?

MichalLytek:
@fmquaglia 
https://github.com/MichalLytek/type-graphql/pull/369#issuecomment-549060921

fmquaglia:
@MichalLytek Thanks. Sorry I wasn't clear. 
I meant "release out of beta"

MichalLytek:
@fmquaglia Soon

PR #393, opened 2019-08-10
feat(fields): add support for nested arrays

### LINKED ISSUE #277: How can I defined a two-dimension array?
Hey, everybody. 
I want to get the value like [[1,2,3], [1,2,3]]. 
What should I do?
```
  @Field(() => [[Int]])
    key?: number[][];
```
It's not work!
It will report:
```
Cannot determine GraphQL output type for key
```
any advice is welcome.
Thank you!!!!

Superd22:
Allow generating nested lists of scalars and object types, such as `[[Int]]`, or `[[CustomObject]]` with mongo like syntax.

```typescript
@ObjectType()
class TestObject {
      // Will become [[String!]!]!
      @Field(type => [[String]])
      NestedArrayField: string[][];
      
      // Will become [[String!]!]
      @Field(type => [[String]], { nullable: true })
      nullableNestedArrayField: string[][] | null;

      // Will become [[String]]!
      @Field(type => [[String]], { nullable: "items" })
      nonNullNestedArrayWithNullableItemField: Array<Array<String | null> | null>;

      // Will become [[String]]
      @Field(type => [[String]], { nullable: "itemsAndList" })
      nestedArrayWithNullableItemField: Array<Array<String | null> | null> | null;
}
```

closes #277

### DIFF
--- a/docs/types-and-fields.md
+++ b/docs/types-and-fields.md
@@ -52,6 +52,8 @@ For simple types (like `string` or `boolean`) this is all that's needed but due 
 - `@Field(type => [Rate])` (recommended, explicit `[ ]` syntax for Array types)
 - `@Field(itemType => Rate)` (`array` is inferred from reflection - also works but is prone to errors)
 
+For nested arrays however, the explicit `[ ]` notation is required to determine the depth of the array. `@Field(type => [[Int]])` would tell the compiler we expect an integer array of depth 2.
+
 Why use function syntax and not a simple `{ type: Rate }` config object? Because, by using function syntax we solve the problem of circular dependencies (e.g. Post <--> User), so it was adopted as a convention. You can use the shorthand syntax `@Field(() => Rate)` if you want to save some keystrokes but it might be less readable for others.
 
 By default, all fields are non nullable, just like properties in TypeScript. However, you can change that behavior by providing `nullableByDefault: true` option in `buildSchema` settings, described in [bootstrap guide](./bootstrap.md).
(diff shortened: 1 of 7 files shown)


19majkel94:
Thank you! ❤️ 
I've made some tweaks in the code to make it simpler and more readable.

We also need an entry in the changelog and update in docs about this syntax (+remove wrong answer from FAQ). Can you do that too? 😉

Superd22:
You're most welcome!

Will do tomorrow!

Superd22:
Done!

docs/types-and-fields.md
```diff
@@ -54,12 +54,16 @@ For simple types (like `string` or `boolean`) this is all that's needed but due
 
 Why use function syntax and not a simple `{ type: Rate }` config object? Because, by using function syntax we solve the problem of circular dependencies (e.g. Post <--> User), so it was adopted as a convention. You can use the shorthand syntax `@Field(() => Rate)` if you want to save some keystrokes but it might be less readable for others.
 
+For nested arrays however, the explicit `[ ]` notation is required to determine the depth of the array. `@Field(type => [[Int]])` would tell the compiler we expect an integer array of depth 2.
```
19majkel94: This should be located paragraph earlier, just after "we have two options available" explanation :)

19majkel94:
Thanks! ❤️ Just one fix and an entry in the changelog... and we are ready to merge 💪

docs/types-and-fields.md
Superd22: Done

19majkel94:
@Superd22 Thank you very much ❤️

Issue #246, opened 2019-02-01
@If condition decorator


Serabass:
What about to add `@If()` decorator?
It must define a field or resolver only if condition is `true`.

For example:

```js
@Resolver(() => ProductMutation)
export class ProductMutation {
    constructor(
        @Inject(Order) private readonly orderRepository: Repository<Order>,
    ) {
    }

    @If(() => process.env.DEFINE_HIDDEN_METHODS === "true")
    @FieldResolver((type) => Number])
    public async hiddenMethod(@Ctx() ctx) {
        return; // ...
    }

    @Authorized()
    @FieldResolver((type) => [Order])
    public async createOrder(): Promise<Order> {
        return this.orderRepository.create();
    }
}
```

19majkel94:
It won't remove this field/method from TS types = not a typesafe solution. I am planing to cover this need in the future but it's a complicated feature.

You can create easily your own `If` decorator if you only need to hide field from graphql schema conditionaly:

```ts
function If(condition: boolean, decorator: MethodDecorator): MethodDecorator {
  return (...args) {
	if (condition) {
      return decorator(...args);
    }
  }
}
```

```ts
    @If(
      () => process.env.DEFINE_HIDDEN_METHODS === "true",
      FieldResolver((type) => Number])
    )
    public async hiddenMethod(@Ctx() ctx) {
        return; // ...
    }
```

Serabass:
I tried it. It works! Thanks!

PR #413, opened 2019-09-02
New Option for Resolvers: Skip schema emit

Xzas22:
I ran in to a situation where I wanted a certain resolver to not emit to the schema during the buildSchema command. 

Specifically, I created a factory resolver class, and for certain instantiations, I want delete${item} to be emitted, and for others I do not. 

Having the ability to mark resolver functions as "skipSchemaEmit: True" would be majorly helpful. 

I added in the necessary options and tests, and confirmed everything works as intended. 

BTW: I am still pretty new to coding in general so my tests are not what I would call "elegant"

### DIFF
--- a/src/decorators/types.ts
+++ b/src/decorators/types.ts
@@ -58,6 +58,9 @@ export interface SchemaNameOptions {
 export interface AbstractClassOptions {
   isAbstract?: boolean;
 }
+export interface SkipSchemaEmitOptions {
+  skipSchemaEmit?: boolean;
+}
 export interface ResolveTypeOptions<TSource = any, TContext = any> {
   resolveType?: TypeResolver<TSource, TContext>;
 }
@@ -65,7 +68,8 @@ export type BasicOptions = DecoratorTypeOptions & DescriptionOptions;
 export type AdvancedOptions = BasicOptions &
   DeprecationOptions &
   SchemaNameOptions &
-  ComplexityOptions;
+  ComplexityOptions &
+  SkipSchemaEmitOptions;
 
 export interface EnumConfig {
   name: string;
--- a/src/helpers/resolver-metadata.ts
+++ b/src/helpers/resolver-metadata.ts
@@ -32,5 +32,6 @@ export function getResolverMetadata(
     description: options.description,
     deprecationReason: options.deprecationReason,
     complexity: options.complexity,
+    skipSchemaEmit: options.skipSchemaEmit,
   };
 }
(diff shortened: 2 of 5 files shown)


Xzas22:
This is my first ever pull request so if I did something wrong please let me know. Im always open to advice on getting better, Thanks!

MichalLytek:
If you need to skip emitting in schema, just comment the decorator 😄 
This kind of API is really awkward 😕

Xzas22:
Its for use with abstract classes. For example:

The factory:
`
export const createBaseResolvers = <E extends AbstractEntity>(
	name: string,
	entity: E,
	inputType: Partial<E> | DeepPartial<E>,
	emitDelete: boolean = false
) => {
	@Resolver(() => entity, { isAbstract: true })
	abstract class BaseResolver {
		protected entity: Repository<E> = entity as any;

		@Query(() => entity, {
			name: `get${name}`,
			description: `Returns a single ${name}`,
			skipSchemaEmit: emitDelete
		})
		async getOne(@Arg(`${name}Id`) id: number): Promise<E | undefined> {
			return await this.entity.findOne(id);
		}

//.... others
`

The instantiation:
`
const BaseCompany = createBaseResolvers('Company', Company, CreateCompanyInput, True);

@Resolver(() => Company)
export class CompanyResolvers extends BaseCompany {
//...whatever other resolvers I want
}
`

This way I have one factory function that I extend from, and when I need to have certain resolvers from that factory on or off I simply add the necessary "TRUE" to their respective skipSchemaEmit option.

MichalLytek:
You can still do that just by not calling the decorator factory:
```ts
const Skip: ClassDecorator & PropertyDecorator & MethodDecorator & ParameterDecorator = () => {};
```

```ts
abstract class BaseResolver {
  @(emitDelete ? Query(() => entity, {
    name: `get${name}`,
    description: `Returns a single ${name}`,
  } : Skip)
  async getOne(@Arg(`${name}Id`) id: number): Promise<E | undefined> {
    return await this.entity.findOne(id);
  }
}
```

Your use case is super rare and I'm not convinced about it 😕

MichalLytek:
I found my earlier solution for skipping decorators, less ugly than the `Skip` :wink:
https://github.com/MichalLytek/type-graphql/issues/246#issuecomment-459717235

```ts
@Resolver(() => entity, { isAbstract: true })
abstract class BaseResolver {
  @If(
    !emitDelete,
    Query(() => entity, {
      name: `get${name}`,
      description: `Returns a single ${name}`,
    }),
  )
  async getOne(@Arg(`${name}Id`) id: number): Promise<E | undefined> {
    return await this.entity.findOne(id);
  }
}
```

But I still think that you should use the class mixins to compose the base resolver conditionally.

So closing this as out of scope as it leads to less readable code 🔒

PR #416, opened 2019-09-08
Add `ClientError` class

lffg:
First of all, thanks for this awesome framework! :)

---

This pull requests creates a new error: `ClientError`, which will be extended by all other error classes that will be sent to the final user.

Why should bother creating this new error?

Well, it is a very common pattern to create an error formatter in our GraphQL APIs. Something like that:

```ts
  // All that good stuff.
  formatError: (error: GraphQLError) => {
    // Out error formatter logic.
  }
});
```

But it is also very common to create "hidden" errors for some internal errors, like a possible schema error in production. So, we tend to do something like:

```ts
function formatError({ originalError }: GraphQLError) {
  if (
    originalError instanceof ApolloError ||
    originalError instanceof MyCustomClientError
  ) {
    return originalError; // No sensitive error data here!
  }

  const id = uuid();
  log({ ...originalError, id });

  return {
    message: 'Internal error',
    id
  }
}
```

So that error messages that could compromise the back-end are hidden by default.

---

But using TypeGraphQL, we can't do this, because some errors are "public"-like and some others don't.

By exposing this `ClientError`, we can permit that kind of check seamlessly.

Thanks o/


MichalLytek:
> because some errors are "public"-like and some others don't

Some maybe export the errors you want to catch, instead of providing another class in the hierarchy? 😕 

> we can permit that kind of check seamlessly

For some users it might help, for others not, e.g. someone wants to hide AuthorizationError as it may leak some info.

There's no need to create an error class to check instance instead of directly four errors. The proposed solution is solving your use case, not a whole range of problem, so I can't accept it.

lffg:
I see... Thanks for your feedback and the nice framework, though. :)

MichalLytek:
@lffg Go ahead and make a PR that exports missing error if you found something :wink:

PR #419, opened 2019-09-16
@types/graphql is included with graphql

Brampage:
In the beginners guide: "installation", it says you need to install @types/graphql. After installing it says: `npm WARN deprecated @types/graphql@14.5.0: This is a stub types definition. graphql provides its own type definitions, so you do not need this installed.`

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -13,7 +13,7 @@ Before getting started with TypeGraphQL we need to install some additional depen
 First, we have to install the main package, as well as [`graphql-js`](https://github.com/graphql/graphql-js) (and it's typings) which is a peer dependency of TypeGraphQL:
 
 ```sh
-npm i graphql @types/graphql type-graphql
+npm i graphql type-graphql
 ```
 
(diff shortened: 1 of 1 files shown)


MichalLytek:
Right, thanks! `graphql-js` since v14.5.0 has built-in TS types definitions.
I've updated peer deps in `v0.18 beta` so we should also update the docs 😉

glen-84:
The live site is still not updated. Also, you may want to remove the `(and it's typings)` text as well.

MichalLytek:
> The live site is still not updated. 

https://typegraphql.ml/docs/next/installation.html

> you may want to remove the (and it's typings) text as well.

Right! Can you do a PR with that change? 😉

glen-84:
> https://typegraphql.ml/docs/next/installation.html

Sure, but that doesn't help new users who are reading the current documentation. 🙂

> Right! Can you do a PR with that change? 😉

https://github.com/MichalLytek/type-graphql/pull/482

Issue #433, opened 2019-10-01
@UseMiddleware( ), Middleware; first argument is undefined


tajpouria:
[ Ubuntu 18.6, node 10.16.3] : I'm just tried to implement basic middleware provided on type-graphql [docs](https://typegraphql.ml/docs/middlewares.html) , but I'm keep receiving following error, is it a problem around my code or a bug withing mentioned library?

http://localhost:4000/graphql
```graphql
# QUERY
query Hello {
  hello
}

# RESPONSE : 
{
  "errors": [
    {
      "message": "Cannot read property 'info' of undefined",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "hello"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "TypeError: Cannot read property 'info' of undefined",
            "    at new exports.ResolveTime (/home/tajpouria/.pro/fun/TypeGraphql-Series/src/modules/user/middleware/logger.ts:3:51)",
            "    at DefaultContainer.get (/home/tajpouria/.pro/fun/TypeGraphql-Series/node_modules/type-graphql/dist/utils/container.js:15:51)",
            "    at IOCContainer.getInstance (/home/tajpouria/.pro/fun/TypeGraphql-Series/node_modules/type-graphql/dist/utils/container.js:36:42)",
            "    at /home/tajpouria/.pro/fun/TypeGraphql-Series/node_modules/type-graphql/dist/resolvers/helpers.js:67:67",
            "    at Generator.next (<anonymous>)",
            "    at /home/tajpouria/.pro/fun/TypeGraphql-Series/node_modules/tslib/tslib.js:110:75",
            "    at new Promise (<anonymous>)",
            "    at Object.__awaiter (/home/tajpouria/.pro/fun/TypeGraphql-Series/node_modules/tslib/tslib.js:106:16)",
            "    at dispatchHandler (/home/tajpouria/.pro/fun/TypeGraphql-Series/node_modules/type-graphql/dist/resolvers/helpers.js:54:28)",
            "    at Object.<anonymous> (/home/tajpouria/.pro/fun/TypeGraphql-Series/node_modules/type-graphql/dist/resolvers/helpers.js:82:16)"
          ]
        }
      }
    }
  ],
  "data": null
}
```

./modules/user/middleware/ResolveTime.ts
```typescript
import { MiddlewareFn } from "type-graphql";

[...]

PR #434, opened 2019-10-02
Update wrong TypeScript compilation target

tajpouria:
Related to issue #433


MichalLytek:
Typescript compilation target setting is described in installation section already:
https://typegraphql.ml/docs/installation.html#typescript-configuration

So there is no need for duplicating this information of configuration in the middlewares docs section.

PR #435, opened 2019-10-05
[fix] ResolverInterface can return null

lookfirst:
if `@FieldResolver({nullable: true})` is possible, then shouldn't ResolverInterface allow a null response?

### DIFF
--- a/src/interfaces/ResolverInterface.ts
+++ b/src/interfaces/ResolverInterface.ts
@@ -3,5 +3,5 @@
  * to provide a proper resolver method signatures for fields of T.
  */
 export type ResolverInterface<T extends object> = {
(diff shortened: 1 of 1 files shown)


MichalLytek:
`ResolverInterface<TObjectType>` is for creating field resolvers for existing `TObjectType` fields, like resolving db relations. So it checks if the "overwritten" field has correct signature - has `TObjectType` on first arg position (`@Root`) and the return type of the method is matching the type of the field.

So if you have `@FieldResolver({nullable: true})`, your field in `TObjectType` also need to have `@Field({nullable: true}) myField?: MyType | null`, so the `ResolverInterface` will allow for returning null from the method.

lookfirst:
Dude, thank you! By the way, absolutely loving your project. It has made working with graphql infinitely more enjoyable. Thank you so much for all the hard work.

Closing as unnecessary.

PR #445, opened 2019-10-16
Add example of how to use with InversifyJS container

evandroabukamel:
*What?*
First, I renamed using-container to using-container-typedi just to show that it uses TypeDI;
Second, I added using-container-inversify folder under examples with a basic example of how to use TypeGraphQL with InversifyJS container and injections;

*Why?*
I propose to add a new example that show how to use InversifyJS with TypeGraphQL. Today at the company I work we have an API written in TypeScript and it uses InversifyJS to implement the Clean Architecture. We used TypeGraphQL before in other projects without InversifyJS , but when we decided to use it I had some problems to put the resolver in the container and then to generate the schema from it. After a lot of researches and fails I finally found this solution that worked for me. I'd like to share it to show that it is possible to use TypeGraphQL with InversifyJS .

*How?*
I implemented the same example used with TypeDI but made minor adaptations and improvements to work with InversifyJS (https://github.com/inversify/InversifyJS).
The tricky part is to put the resolver in the container, because we have to use the "self-binding of concrete types" (https://github.com/inversify/InversifyJS/blob/master/wiki/classes_as_id.md#self-binding-of-concrete-types). I tried several approaches and different ways of binding but only doing this that everything worked fine.

I believe this will help other developers sometime because I could not find such a clear information about this anywhere. If you have some questions or comments, I'll be willing to answer.


MichalLytek:
> I believe this will help other developers sometime because I could not find such a clear information about this anywhere.

I think that a better place for such info:

> The tricky part is to put the resolver in the container, because we have to use the "self-binding of concrete types" (inversify/InversifyJS:wiki/classes_as_id.md@master#self-binding-of-concrete-types ). I tried several approaches and different ways of binding but only doing this that everything worked fine.

would be a paragraph in "Dependency injection" docs page. It makes no sense to duplicate the example just to show a few lines of required config 😕

evandroabukamel:
I agree with you. I would be nice this In the documentation but the example could be very helpful too. The docs folder in the repository are the same on the website?

MichalLytek:
> The docs folder in the repository are the same on the website?

Sure, the website is generated from *.md docs files.

> the example could be very helpful too

If somebody knows and use Inversify, the info about requirement to self-bind resolver classes will be enough 😉

evandroabukamel:
Ok, I will make the change as soon as I can.

evandroabukamel:
I created another PR for the documentation, https://github.com/MichalLytek/type-graphql/pull/447.
We can close this you prefer.

PR #447, opened 2019-10-20
docs(container): add InversifyJS side note

evandroabukamel:
Added example:
- add an example in the 'Dependency injection' page that shows the proper way to bind a resolver class if InversifyJS is being used.

The reason I've done this is because I had some trouble to get it working, it wasn't easy to gather information about Inversify and GraphQL, I tried different approaches and that's the only one that worked for me.


MichalLytek:
@evandroabukamel 
I've refactored that to a side note as it's not important for everyone 😉

evandroabukamel:
Nice, thank you.

PR #452, opened 2019-10-29
POC: nested input support

j:
This is an attempt at fixing inputs.

I'm doing this by iterating over the input type fields, checking if the field's type is an input type, and if so, hydrating the argument into the class.  I attempted to make it more performant by caching the tree based off the param metadata's type.


MichalLytek:
@j Your idea is good! 💪 I will try to work on this to support all cases arrays, scalars and nulls.

MichalLytek:
Closing in favor of #462 🔒

PR #462, opened 2019-11-10
Fix transforming and validating nested inputs and arrays

### LINKED ISSUE #133: Nested input types are not deserialized into class instances
**Describe the bug**
When having nested input types only the top-most class gets transformed into it's instance. Any nested objects remains plain javascript objects.

**To Reproduce**
```ts
@InputType()
class NestedTestInputType {
  @Field()
  foo: number
}

@InputType()
class TestInputType {
  @Field()
  nested: NestedTestInputType
}

@Service()
@Resolver(of => /* ... */)
export class TestResolver {
   @Query(returns => [String])
  async test(@Arg('filter') filter: TestInputType) {
    assert(filter instanceof TestInputType, 'Filter must be instance of TestInputType') // <-- OK
    assert(
      filter.nested instanceof NestedTestInputType,
      'Filter#nested must be instance of NestedInputType',
    ) // <-- FAILS
    return []
  }
}
```

**Expected behavior**
Nested input types should be deserialized into instances of their classes recursively.

- type-graphql@0.12.3

MichalLytek:
Closes #133 🔒

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - add basic support for directives with `@Directive()` decorator (#369)
 ### Fixes
 - refactor union types function syntax handling to prevent possible errors with circular refs
+- fix transforming and validating nested inputs and arrays (#462)
 
 ## v0.17.5
 ### Features
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -98,6 +98,8 @@ Note that by default, the `skipMissingProperties` setting of the `class-validato
 
 GraphQL will also check whether the fields have correct types (String, Int, Float, Boolean, etc.) so we don't have to use the `@IsOptional`, `@Allow`, `@IsString` or the `@IsInt` decorators at all!
 
+However, when using nested input or arrays, we always have to use [`@ValidateNested()` decorator](https://github.com/typestack/class-validator#validating-nested-objects) or [`{ each: true }` option](https://github.com/typestack/class-validator#validating-arrays) to make nested validation work properly.
+
 ## Response to the Client
 
 When a client sends incorrect data to the server:
--- a/src/helpers/types.ts
+++ b/src/helpers/types.ts
@@ -100,6 +100,10 @@ export function convertToType(Target: any, data?: object): object | undefined {
   if (data instanceof Target) {
     return data;
   }
+  // convert array to instances
+  if (Array.isArray(data)) {
+    return data.map(item => convertToType(Target, item));
+  }
 
   return Object.assign(new Target(), data);
 }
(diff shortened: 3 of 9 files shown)


MichalLytek:
@j Please review the changes - they are based on your #452 PR 😉

j:
@MichalLytek looks like familiar code! :P haha but looks great.  I'll get to playing with it soon.  I moved forward with a lot of internal development and was going to add validation at some point later.  Thanks for this.

MichalLytek:
@j Soooo have you checked the changes? Does it work with edge use cases? 😄

j:
@MichalLytek woops, I wrote that on my phone and thought it was merged.  I'll test this now.

j:
@MichalLytek based on our current code-base where we use inheritance on the root level of an input and nested inputs, I was able to confirm that validation works and all inputs are mapped correctly.  I toyed with doing deep nested with inheritance and those were converted properly as well.

Thanks for this!

j:
@MichalLytek one thing I just saw, and I'm not sure if this is a bug or not.  If you have a nullable input, it's creating the object with undefined fields.

j:
Extra info:

The below mutation has a nullable input, in this case, I left it out and errors happen.

```graphql
mutation createCart {
  createCart {
     ...
  }
}
```

```
[
            "TypeError: Cannot read property 'email' of undefined",
            "    at tree.getFields.reduce (/Users/graphql/node_modules/type-graphql/dist/resolvers/convert-args.js:54:27)",
            "    at Array.reduce (<anonymous>)",
            "    at convertToInput (/Users/graphql/node_modules/type-graphql/dist/resolvers/convert-args.js:52:42)",
            "    at convertValueToInstance (/Users/graphql/node_modules/type-graphql/dist/resolvers/convert-args.js:71:11)",
            "    at convertValuesToInstances (/Users/graphql/node_modules/type-graphql/dist/resolvers/convert-args.js:78:12)",
            "    at Object.convertArgToInstance (/Users/graphql/node_modules/type-graphql/dist/resolvers/convert-args.js:104:12)",
            "    at Object.<anonymous> (/Users/graphql/node_modules/type-graphql/dist/resolvers/helpers.js:17:76)",
            "    at Generator.next (<anonymous>)",
            "    at /Users/graphql/node_modules/tslib/tslib.js:110:75",
            "    at new Promise (<anonymous>)"
          ]
```

MichalLytek:
Thanks for the report, @j! That's the use cases I was thinking of 😄 

I've checked the nullable fields but they are handled by `convertToType`.
But the if the value of nullable arg `@Arg("input", { nullable: true }) input?: SampleNestedInput` is null, then it failed, so I've fixed it 💪

j:
@MichalLytek yup, that ended up fixing the error! 💪

MichalLytek:
@j Thanks for confirmation!

Merging 👀

Issue #350, opened 2019-05-31
browser.shim.js error for internet explorer 11


domi7777:
**Describe the bug**
`browser.shim.js` cannot be processed by internet explorer 11 because it contains arrow function code: 

`const dummyFn = () => void 0;`

I was able to fix the issue by changing it to:
```
var dummyFn = function () {
  return function () {
  };
};
```

Now, I'm not sure if the `browser.shim.ts` should be updated likewise, or if typescript compilation target could be set to ecmascript 5 for this file only.

19majkel94:
@domi7777 
TypeGraphQL is a Node.js framework and it requires ES2016 target to work because it uses a few new features of EcmaScript:
https://typegraphql.ml/docs/installation.html#typescript-configuration

So it makes no sense to transpile the lib to ES5. What is your use case that you need IE 11 support?

domi7777:
@19majkel94 , as documented in https://typegraphql.ml/docs/browser-usage.html , we use the browser.shim to be able to share our typescript model in the frontend, so the browser.shim.js ends up in the browser.

19majkel94:
@domi7777 And your bundler doesn't handle transpilation to the target ES version?

domi7777:
@19majkel94 our tsconfig.json contains   `  "target": "es5",` already, and we don't compile browser.shim.ts as it is already transpiled to browser.shim.js in the `node_modules/type-graphql`.

19majkel94:
@domi7777
But a lot of packages in npm are not distributed in ES5-compatible version, so you need to transpile them too to make it work in IE11 or older browser.

domi7777:
@19majkel94 I guess I could add babel to my build to transpile this file, but I thought it would be overkill.

The other libs I use provide different versions of their scripts (umd/esm5/esm2015/...)

19majkel94:
@domi7777 
Can you point me to the example lib that has multiple entrypoints for different target/build systems? I would like to support them too.

domi7777:
@19majkel94 I'm really no expert on this matter, but here's an example: [apollo-angular](https://github.com/apollographql/apollo-angular). 

I'm not sure it is worth investing time to support fully Ecmascript5 as type-graphql is a server-side library, and only the browser.shim.js is used on frontend side as a way to not include the library in the bundle. 

That is why I thought simply changing the two arrow functions to es5 functions would make internet explorer happy without the burden of changing the build configuration.

19majkel94:
[...]

PR #466, opened 2019-11-18
Fix browser shim TypeScript argument checking

austin-beer:
Some of the decorators in this library take optional arguments. Thus the
shimmed version of the decorator factories should optionally accept
arguments as well.

Also, as documented by
https://www.typescriptlang.org/docs/handbook/decorators.html, the
functions returned by decorator factories may take arguments. Thus the
shimmed version of these functions should optionally accept arguments as
well.

### DIFF
--- a/src/browser-shim.ts
+++ b/src/browser-shim.ts
@@ -12,8 +12,8 @@ plugins: [
 ]
 */
 
-const dummyFn = () => void 0;
-const dummyDecorator = () => dummyFn;
+const dummyFn = (...args: any[]) => void 0;
+const dummyDecorator = (...args: any[]) => dummyFn;
 
 export const Arg = dummyDecorator;
 export const Args = dummyDecorator;


MichalLytek:
They are called with an argument, but you can just not use them and it's perfectly legal. So you don't have to create params signature if you don't use them:

```ts
type MyFn = (a: string, b: number, c: boolean) => void;

const myFn: MyFN = () => void 0;
```

So this change makes no sense as it's designed to replace the JS version in runtime only.

austin-beer:
Agreed. The issue is with Typescript's function signature validation using declarations from the generated browser-shim.d.ts file. With the existing browser shim code, I get TypeScript errors when transpiling the browser shim in a TypeScript project. For example, TypeScript says I'm providing 2 arguments to `@Field()` when it expected 0 when I use `@Field(type => ID, {nullable: true})`.

austin-beer:
An alternative solution which would fix both this issue and #350 would be to replace browser-shim.ts in the source directory with hand-coded browser-shim.js and browser-shim.d.ts that are simply copied over to the output directory and packaged as-is. Then browser-shim.js could be changed to not use arrow functions (which would fix #350) and browser-shim.d.ts could be changed to accept all function signatures (which would fix this issue).

If you'd like me to update this PR with that approach, let me know.

MichalLytek:
I still don't understand what is the issue.

Browser shim is designed to replace the `type-graphql` whole bundle with a lightweight mock that just allow to reuse your classes with e.g. class-validator decorators on your client. It should be done on the bundle phase, like webpack or rollup.

This shim is not intended to be directly imported in your code - you should only use the `*.js` file to replace the runtime by the bundler.

austin-beer:
Ah, I see where you're coming from. I should have mentioned that I'm not using webpack or rollup directly. I'm using Angular CLI (ng) to serve and build my UI application. So I'm pulling in the browser shim by adding the following to my tsconfig.json file.
```
{
  "compilerOptions": {
    "baseUrl": "src",
    "paths": {
      "type-graphql": [
        "../node_modules/type-graphql/dist/browser-shim"
      ]
    }
  }
}
```
So the changes in this PR allow the browser shim to be imported directly into the UI code. I think that's still a beneficial change.

MichalLytek:
Does Angular respects TypeScript paths mapping and properly transform the imports while building the app?

Maybe you should try to point to the `.js` file so it will ignore the typings?

austin-beer:
Yes, I believe Angular properly transforms the imports while building the application. At least, it is for me.

I can change it to point to the .js file and it ignores the typings and everything works well, _if_ TypeScript's strict mode is off. But if it is on, like it is in my project, then I get errors about not being able to find a declaration file for module 'type-graphql'. Supporting TypeScript's strict mode seems desirable to me.

MichalLytek:
But with this fix it will still yell about implicit any in callbacks `@Field(type => MyType)`, etc.

Using the real types signatures might help but I'm afraid it will leak and affect the size of the shim 😕 Have to check that approach.

austin-beer:
Yes, I did have to change my `@Field(type => MyType)` to `@Field((type: any) => MyType)`, but other than that I didn't have to change anything else to satisfy TypeScript's strict mode.

I don't believe that `.d.ts` files get packaged normally, only the `.js` files, and this PR isn't creating any new dependencies in that file, so it shouldn't affect the size of the shim.

austin-beer:
Superseded by #469.

PR #469, opened 2019-11-21
Add support for the browser shim in Angular

### LINKED ISSUE #350: browser.shim.js error for internet explorer 11

austin-beer:
This supersedes #466 and #467.

Further testing revealed that my previous PRs didn't work with Angular's AoT compiler. This solution fully supports Angular/TypeScript UI projects using browser-shim.ts while keeping support for non-TypeScript UI projects using browser-shim.js (and making that ES5-compatible).

This also should fix #350.

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -67,6 +67,21 @@ export class Gulpfile {
   }
 
   /**
+   * Copy over the browser-shim files instead of compiling them.
+   * The .js file supports ES5 and newer browsers.
+   * The .ts file supports Angular and other TypeScript projects.
+   */
+  @MergedTask()
+  packageCopyBrowserShim() {
+    return [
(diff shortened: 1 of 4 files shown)


MichalLytek:
@austin-beer 
I've created a #473 with type safety features, different approach of gulp script and a upgraded documentation. Please check if that works well with Angular so I can merge it 💪

austin-beer:
Superseded by #473

PR #472, opened 2019-11-23
Update docs to use Apollo Server plugins for dependency injection

rocketeerbkw:
Since 2.2.0, Apollo Server has plugin support which can be used to clean up scoped containers. I haven't tested this method with `type-graphql` but I used it to clean up database connections.


MichalLytek:
@rocketeerbkw 

I've changed the code the examples and tested that it works well 💪 
And as you had a few typos in the text, I've rewrote it on my own and also include it with the latest stable docs as it's not related to TypeGraphQL version.
https://typegraphql.ml/docs/dependency-injection.html#scoped-containers

PR #473, opened 2019-11-23
Upgraded browser shim

MichalLytek:
This PR adds TS-compatible browser shim that is required for Angular's project with AOT enabled.
It also fixes compatibility with ES5 environment.

Closes #469, #350.

### DIFF
--- a/docs/browser-usage.md
+++ b/docs/browser-usage.md
@@ -9,12 +9,28 @@ Sometimes we might want to use the classes we've created and annotated with Type
 Since TypeGraphQL is a Node.js framework, it doesn't work in a browser environment, so we may quickly get an error, e.g. `ERROR in ./node_modules/fs.realpath/index.js`, while trying to build our app with Webpack. To correct this, we have to configure Webpack to use the decorator shim instead of the normal module. We simply add this plugin code to our webpack config:
 
 ```js
-plugins: [
-  // ...here are any other existing plugins that we already have
-  new webpack.NormalModuleReplacementPlugin(/type-graphql$/, resource => {
-    resource.request = resource.request.replace(/type-graphql/, "type-graphql/dist/browser-shim");
-  }),
-];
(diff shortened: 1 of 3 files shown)


austin-beer:
I've tested this and it works without issue in my Angular UI project. Thanks!

nphias:
@austin-beer @MichalLytek could you give an account of how to make type-graphql work with angular 8.x .?  I am using apollo ... and a little confused on how to buildschema from my app.module as well as integrate resolvers + type objects into a service class. I was unable to find any examples in github/google

MichalLytek:
@nphias I don't use Angular so I can't help you with that.
There is an `apollo-client` example about using resolvers on client side, that's all.
The browser shim is just for sharing object types and input types in client without bundling whole library.

austin-beer:
@nphias, we use Angular on the client side and Node.js on the server side. We only really use type-graphql on the server side where we define our resolvers. However, the type-graphql ObjectType classes and enums that we create on the server side we use in our Angular client. That's where we use the type-graphql browser shim to ignore all of the type-graphql decorators that the Angular client doesn't know what to do with. In our Angular tsconfig.json we have the following:

```
{
  "compilerOptions": {
    "baseUrl": "src",
    "paths": {
      "type-graphql": [
        "../node_modules/type-graphql/dist/browser-shim"
      ]
    }
  }
}
```

nphias:
yeah thanks for your responses  @MichalLytek and @austin-beer i am not really doing client-server graphql... only a one machine hybrid..  i want to use angular.. and type-graphql together.. so its a question of wrapping things like the resolver object into a angular injectable service class and injecting my own connectionManager service into that..   the makeschema asks for resolver object so i guess i am going to try and inject the resolver server class into my angular graphql module and provide a method call.....   i am a die hard OOP guy.. i like angular....  FP everything bothers me..

Issue #480, opened 2019-12-04
typegraphql-prisma gets stuck installing


codejamninja:
`typegraphql-prisma` gets stuck installing.

I think it's related to https://github.com/prisma/photonjs/issues/310

I think updating the prisma dependencies will fix it.

MichalLytek:
> However, during the preview I would ask to add comments only in this issue, not to create a new issues, as I don't want to pollute the repo for now 😉

https://github.com/MichalLytek/type-graphql/issues/476

codejamninja:
@MichalLytek sure

PR #481, opened 2019-12-04
Updated prisma

codejamninja:
Fixes issue #480


MichalLytek:
Unfortunately it's not fixed. And I will update the deps manually with checking for breaking changes and new features. Thanks anyway! 😉

Issue #254, opened 2019-02-14
Too slow for simple large list


sixmen:
**Is your feature request related to a problem? Please describe.**
type-graphql is slower than normal graphql for simple large list.

type-graphql: `time = 920`
```
import 'reflect-metadata';
import { ObjectType, Field, Resolver, Query, buildSchemaSync } from 'type-graphql';
import { graphql } from 'graphql';

@ObjectType()
class Item {
  @Field()
  a: number;
  @Field()
  b: number;
  @Field()
  c: number;
  @Field()
  d: number;
  @Field()
  e: number;
}

@Resolver(Item)
class ItemResolver {
  @Query(() => [Item])
  items() {
    return new Array(5000).fill(0).map((v, i) => ({ a: i, b: i, c: i, d: i, e: i }))
  }
}

const schema = buildSchemaSync({
  resolvers: [ItemResolver]
});

const start = Date.now();
graphql(schema, '{ items { a b c d e } }')
  .then((r) => {
    console.log(r.data.items.length);
    console.log(r.data.items[30].c);
    console.log('time =', Date.now() - start);
  });
```

normal graphql: `time = 101`
```
import {
  graphql,
  GraphQLSchema,
  GraphQLObjectType,
  GraphQLString,
  GraphQLList
} from 'graphql';

const Item = new GraphQLObjectType({
  name: 'Item',
  fields: {
    a: { type: GraphQLString },
    b: { type: GraphQLString },
    c: { type: GraphQLString },
    d: { type: GraphQLString },
    e: { type: GraphQLString },
  }
});

var schema = new GraphQLSchema({
  query: new GraphQLObjectType({
    name: 'RootQueryType',
    fields: {
      items: {
        type: GraphQLList(Item),
        resolve() {
          return new Array(5000).fill(0).map((v, i) => ({ a: i, b: i, c: i, d: i, e: i }))
        }
      }
    }
  })
});

const start = Date.now();
graphql(schema, '{ items { a b c d e } }')
  .then((r) => {
    console.log(r.data.items.length);
    console.log(r.data.items[30].c);
    console.log('time =', Date.now() - start);
  });
```

**Describe the solution you'd like**
I think some optimizations are needed for simple cases.

If I change createSimpleFieldResolver, it runs faster.

```
function createSimpleFieldResolver(fieldMetadata) {
    return (root, args, context, info) => root[fieldMetadata.name];
}
```

19majkel94:
Note: partially related to #71.

[...]

PR #488, opened 2019-12-10
Optimize resolvers execution paths

### LINKED ISSUE #487: FieldResolver is too slow
**Is your feature request related to a problem? Please describe.**
This is an advanced topic of #254

If you use FieldResolver, execution needs more time.

```
import 'reflect-metadata';
import { buildSchemaSync, Field, FieldResolver, ObjectType, Query, Resolver, Root } from 'type-graphql';
import { graphql } from 'graphql';

@ObjectType({ simpleResolvers: true })
class Item {
  @Field()
  a: number;
  @Field()
  b: number;
  @Field()
  c: number;
  @Field()
  d: number;
  @Field()
  e: number;
}

@Resolver(Item)
class ItemResolver {
  @Query(() => [Item])
  items() {
    return new Array(5000).fill(0).map((v, i) => ({ a: i, b: i, c: i, d: i, e: i }))
  }

  @FieldResolver()
  a(@Root() source) {
    return source.a;
  }

  @FieldResolver()
  b(@Root() source) {
    return source.b;
  }

  @FieldResolver()
  c(@Root() source) {
    return source.c;
  }

  @FieldResolver()
  d(@Root() source) {
    return source.d;
  }

  @FieldResolver()
  e(@Root() source) {
    return source.e;
  }
}

const schema = buildSchemaSync({
  resolvers: [ItemResolver]
});

const start = Date.now();
graphql(schema, '{ items { a b c d e } }')
  .then((r) => {
    console.log(r.data!.items.length);
    console.log(r.data!.items[30].c);
    console.log('time =', Date.now() - start);
  });
```

**Describe the solution you'd like**
I didn't think whether some option likes simple is a solution or not.

But in the end, some optimization technique applied in graphql-js may be needed, I think.

For example,
```
    if (isPromise(result)) {
      return result.then(undefined, error => {
        exeContext.errors.push(error);
        return Promise.resolve(null);
      });
    }
    return result;
```

(from https://github.com/graphql/graphql-js/blob/master/src/execution/execute.js)

In graphql-js, those two declarations become difference result.

```
const Item = new GraphQLObjectType({
  name: 'Item',
  fields: {
    a: { type: GraphQLString, resolve: (s) => s.a },
  }
});

const Item = new GraphQLObjectType({
  name: 'Item',
  fields: {
    a: { type: GraphQLString, resolve: async (s) => s.a },
  }
});
```

sixmen:
Fixes #487


MichalLytek:
@sixmen 
Please add proper entries to the results in the benchmarks folder - without measuring it makes no sense to add any "optimization".

MichalLytek:
> So, I think it is still useful to optimize Promises.
> https://github.com/MichalLytek/type-graphql/issues/487#issuecomment-564042385

You think that but I've measured that it's not useful - the overhead of async/await is unmeasurable. 

I've checked with the `undefined` resolver (implicit one), the explicit sync (`root => root[field.name]`) and an async one (`async root => await root[field.name]`).

Try to guess which results comes from which case 😉 

- 12018.493ms
- 11839.115ms
- 12207.592ms
- 11822.750ms
- 11719.892ms

So this kind of "optimization" makes no sense and I have to reject this.

Field resolvers are a standard resolvers and needs full execution model like params injection, custom decorators, middlewares, args validation, class instance, etc. This is the real overhead, not the promises, unless you use Node 12 LTS or newer, not the unmaintained soon Node 8. 

Detecting which feature is used when is too complicated for now so it's out of scope as they are many other tasks waiting in the queue.

> (maybe no need for simple option?)
> https://github.com/MichalLytek/type-graphql/issues/487#issuecomment-564042385

Simple option allows to skip global middlewares so it's useful if you want to trade-off field access control/logging for performance, or strictly to opt-in for better performance, e.g. when using Typegoose middleware. 

I think that all we can do is to add `{ raw: true }` support for full resolvers like `@Query` or `@FieldResolver` that will skip all the TypeGraphQL things and executes using a standard `(root, args, ctx, info)` signature. What do you think @sixmen ?

MichalLytek:
Lesson 1: always build your code before running 😄 

```
Sync resolver:  12118.231ms
Async/await:    28740.372ms
```

So this makes sense to bloat the code with the fast sync path 🤔

sixmen:
I also don't like unnecessary optimization that ruins the code.

(It may be unreal cases, but I faced the case as #254 shown.
I cannot be sure for FieldResolver.
I faced a problem with async FieldResolver, and I avoided it with no FieldResolver with cached value)

So I tried to minimize the change.

I just made this PR to show you the possibility of optimization.
`{simpleResolvers:true}` was too difficult to find resolvers that need to apply.
 
I added the benchmark commit, and following is the result from my computer (MacBook Pro i7 2.6GHz)

| Benchmark | Node 10.16.0 master | Node 13.3.0 master | Node 10.16.0 PR | Node 13.3.0 PR |
|---|---|---|---|---|
| simple-single-raw | 472.977ms | 408.466ms | 464.077ms | 416.154ms |
| simple-nested-raw | 1305.241ms | 1.270s | 1322.084ms | 1.233s |
| simple-single-tg | 1271.616ms | 745.608ms | 504.383ms | 428.95ms |
| simple-nested-tg | 4614.922ms | 2.222s | 1419.520ms | 1.262s |
| large-basic-raw | 2535.619ms | 2.661s | 2550.349ms | 2.610s |
| large-sync-raw | 2466.865ms | 2.616s | 2518.828ms | 2.625s |
| large-async-raw | 6623.758ms | 6.315s | 6569.861ms | 6.251s |
| large-basic-tg | 20637.726ms | 15.165s | 3192.436ms | 3.143s |
| large-sync-tg | 39585.782ms | 24.437s | 4292.903ms | 4.097s |
| large-async-tg | 40319.222ms | 23.873s | 9920.493ms | 9.594s |

MichalLytek:
I wonder why field resolvers are still 50% slower than the raw `graphql-js` - with `simpleResolvers` it's only 15% 🤔

sixmen:
@MichalLytek do you mean "Allow edits from maintainers."? I enabled it.

I will also look for more bottlenecks, if I had time.

sixmen:
I found another possibility.


In my project, convertToType causes a problem.

An ObjectType class of my project has a read-only field (using Object.defineProperty).
If this object is passed to `@Root`, convertToType will try to `Object.assign(new Target(), data)` and fail.
I want to skip create an object, but I want type checking working.
So I used type for the class by `type MyClassVO = Pick<MyClass, keyof MyClass>`.


If I changes `@Root() source: SampleObject` to `@Root() source: any`, it runs faster.

MichalLytek:
> do you mean "Allow edits from maintainers."? I enabled it.

@sixmen 
image

sixmen:
스크린샷 2019-12-14 오후 1 46 09

I enabled it. I don't know what is the reason.

Instead, I invited you to the forked repository.

MichalLytek:
Fun story - `"then" in value` messing up something in Promises so when I call it in `isPromiseLike`, the tests are not passing 😕

sixmen:
> Fun story - `"then" in value` is messing up something in Promises so when I call it in `isPromiseLike`, the tests are not passing 😕 Reverted to the type assertion

Two are different:

```
> a = 2
2
> a.then
undefined
> 'then' in a
Thrown:
TypeError: Cannot use 'in' operator to search for 'then' in 2
```

More accurate version may be https://github.com/then/is-promise/blob/master/index.js, I copied it from https://github.com/graphql/graphql-js/blob/master/src/jsutils/isPromise.js.

MichalLytek:
Right, both me and TypeScript forgot that the 'in' operator can be used only on objects, not primitives.
But I will stick with the simpler version as it's measurably faster 😄

MichalLytek:
@sixmen 
I am really astonished how things speeds up with this changes.
Now the simplest cases takes 15.518s instead of 57.776s, which makes `simpleResolvers` not needed then. It speeds up even when global middleware are used (80.517s vs 62.664s)!

I've also added a benchmark case when `simpleResolvers` are usefull, as if you have a global middleware or auth checker, it reduces from 62.664s to 14.980s 😉 

The last thing I have to do in this PR is to update the performance docs with this insights.

sixmen:
When you will release this commit? I need to decide whether waiting next release or using beta release.

MichalLytek:
@sixmen 
I am grouping breaking changes before releasing major 0.18 version.
You can safely install `type-graphql@beta` - it's not an experimental branch, it's a master branch that is a release candidate 😉 
I am collecting feedback about this changes as it sometimes breaks others code: #510

PR #515, opened 2020-01-08
check if resolveType returns a promise

benawad:
This is blocking me from using `graphql-jit` atm. It doesn't expect resolveType to return a promise.


benawad:
changed wrong file

PR #516, opened 2020-01-08
check if resolveType returns a promise

benawad:
This is blocking me from using `graphql-jit` atm. It doesn't expect resolveType to return a promise.

### DIFF
--- a/src/schema/schema-generator.ts
+++ b/src/schema/schema-generator.ts
@@ -41,7 +41,13 @@ import {
   ConflictingDefaultValuesError,
   InterfaceResolveTypeError,
 } from "../errors";
-import { ResolverFilterData, ResolverTopicData, TypeResolver } from "../interfaces";
+import {
+  ResolverFilterData,
+  ResolverTopicData,
+  TypeResolver,
+  Maybe,
+  ClassType as ClassTypeType,
+} from "../interfaces";
 import { getFieldMetadataFromInputType, getFieldMetadataFromObjectType } from "./utils";
 import { ensureInstalledCorrectGraphQLPackage } from "../utils/graphql-version";
 import {
@@ -50,6 +56,7 @@ import {
   getInputValueDefinitionNode,
   getObjectTypeDefinitionNode,
 } from "./definition-node";
+import isPromiseLike from "../utils/isPromiseLike";
 
 interface AbstractInfo {
   isAbstract: boolean;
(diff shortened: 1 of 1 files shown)


MichalLytek:
>  It doesn't expect resolveType to return a promise.

1) This is against GraphQL JS runtime behavior:

https://github.com/graphql/graphql-js/tree/b4bff0ba9c15c9d7245dd68556e754c41f263289/tstypes/type/definition.d.ts#L419-L430

2) This doesn't solve the issue, only makes a workaround that it will work until you for some reason return a promise 😕 

It's better to raise an issue to fix in GraphQL-JIT rather that restricting even on `graphql-js` side.

benawad:
> This is against GraphQL JS runtime behavior:

yeah I brought it to their attention.

> This doesn't solve the issue, only makes a workaround that it will work until you for some reason return a promise 😕

Either way, forcing every resolveType resolver to return a promise is worse for performance. Why not let the developer choose whether they want it to return a promise or not?

src/schema/schema-generator.ts
```diff
@@ -663,12 +670,26 @@ export abstract class SchemaGenerator {
...
     resolveType: TypeResolver<TSource, TContext>,
   ): GraphQLTypeResolver<TSource, TContext> {
-    return async (...args) => {
-      const resolvedType = await resolveType(...args);
-      if (typeof resolvedType === "string") {
-        return resolvedType;
+    return (...args) => {
+      const resolvedTypeOrPromise = resolveType(...args);
+
+      if (isPromiseLike(resolveType)) {
```
MichalLytek: I think it should be `resolvedTypeOrPromise` - we are checking the return value of `resolveType` fn

src/schema/schema-generator.ts
```diff
@@ -41,7 +41,13 @@ import {
   ConflictingDefaultValuesError,
   InterfaceResolveTypeError,
 } from "../errors";
-import { ResolverFilterData, ResolverTopicData, TypeResolver } from "../interfaces";
+import {
+  ResolverFilterData,
+  ResolverTopicData,
+  TypeResolver,
+  Maybe,
+  ClassType as ClassTypeType,
```
MichalLytek: Why do you need such alias?

MichalLytek:
> Why not let the developer choose whether they want it to return a promise or not?

As you can see, introducing sync paths results in duplicated code. TypeGraphQL was not designed with this kind of optimization in mind but the vNext is:

https://github.com/MichalLytek/type-graphql/blob/16d482d9537b87f16c95198c52f929f71ae065a9/packages/core/src/runtime/RuntimeGenerator.ts#L25-L46

Sure we can merge this PR but don't expect any refactor or future async-related optimization as I don't have enough free time to actively support two versions of TypeGraphQL.

benawad:
I'm cool waiting for vNext for this change. I just wasn't sure if this was going to be fixed with it.

MichalLytek:
Every "maybe async" execution path will be optimized in vNext for sync cases with `completeValue` helper - I should write a rule "no await/.then()" for the `runtime` folder 😄

PabloSzx:
what happened with this issue @MichalLytek ? right now it's impossible to use unions with TypeGraphQL & Mercurius due to this limitation

MichalLytek:
@PabloSzx it's out of scope - please ask `graphql-jit` team to support and conform to GraphQL JS specification

StarpTech:
GraphQL JS specification says either promise or value. The current direction here is to enforce promises.

PR #521, opened 2020-01-10
Add @Extensions decorator

hihuz:
Hello :wave:

First of all, thank you very much for your work on this library, it has helped us a lot in rapidly creating a graphql endpoint to go hand in hand with our existing REST API 🙂 

In this PR, I will attempt to address this issue: https://github.com/MichalLytek/type-graphql/issues/124

I have tried to base my implementation off of the comments I could read in the issue and also in the PRs that were previously created for this. I also took inspiration from the Directive PR (https://github.com/MichalLytek/type-graphql/pull/369) which helped me by showcasing recent good PR practices on the project.

I will leave several comments of my own where I am unsure about my approach, I would be grateful if you could let me know your views on them, and also for any time you would spend reviewing the PR.

I won't be very active during the upcoming week-end, but I'll definitely try to address any feedback as fast as I can during the next week.

Best!


examples/extensions/authorizer.middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+  const fieldAuthorizationExtensions = (field.extensions && field.extensions.authorization) || {};
+
+  return {
+    ...parentAuthorizationExtensions,
+    ...returnTypeAuthorizationExtensions,
+    ...fieldAuthorizationExtensions,
+  };
+};
+
+export const AuthorizerMiddleware: MiddlewareFn = async (
```
hihuz: I have chosen to base my examples off of the "authorization" examples, and showcase how we could implement it in a custom way with extensions, I hope it makes sense?

examples/extensions/authorizer.middleware.ts
```diff
@@ -0,0 +1,43 @@
+import { GraphQLResolveInfo, GraphQLObjectType } from "graphql";
+
+import { MiddlewareFn } from "../../src";
+import { Context } from "./context.interface";
+import { UnauthorizedError } from "../../src/errors";
+
+const extractAuthorizationExtensions = (info: GraphQLResolveInfo) => {
```
hihuz: I would love to get some feedback on this, I am not so experienced with schema inspection and I am wondering if there is a more elegant way to do this? I was expecting it to be more easy to "read" the meta data from the field being resolved.

The order in which I am overriding the extensions is arbitrary of course.

I am genuinely interested if there is a better way since we will be using this feature for authorization ourselves 🙂

src/metadata/utils.ts
```diff
@@ -57,3 +60,10 @@ export function ensureReflectMetadataExists() {
     throw new ReflectMetadataMissingError();
   }
 }
+
+export function flattenExtensions(
```
hihuz: I am not very happy with this, having an "util" for just a specific field does not make a lot of sense to me but I also didn't want to repeat the reducer function twice.

I can propose two alternatives:
* make a more generic util which will flatten by key passed as argument (it seemed a little bit premature to me since there are no other cases in the codebase yet, but I could do it)
* simply use an inline function duplicated twice and remove this util

What do you think?

src/schema/schema-generator.ts
```diff
@@ -498,6 +502,7 @@ export abstract class SchemaGenerator {
         astNode: getFieldDefinitionNode(handler.schemaName, type, handler.directives),
         extensions: {
           complexity: handler.complexity,
```
hihuz: It seems correct to me that `extensions.complexity` would be overridden by regular extensions, seeing as if I understood it correctly the plan is to refactor `complexity` and other decorators to use `extensions` at some point.

Bottom line: if `@Extensions` is a lower level decorator, it should take precedence over others if users would like to implement custom stuff

MichalLytek: Yes, the complexity decorator can be replaced by the extensions. Preventing colliding properties names is the users responsibility.

src/metadata/utils.ts
MichalLytek: Yes, you should just inline the spread operator usage, it makes no sense to wrap single line in dedicated 5 lines function

hihuz: Thanks for the feedback, I have changed it in this commit: 6eac60ee487040f4072de05259e28539afe9bdf5

Please let me know if you'd rather have me wait until your review is over before I start addressing your comments 🙂

MichalLytek:
@hihuz 

image

image

hihuz:
@MichalLytek Hmm my mistake 🙁I did enable this option but it seems our organization rules overrides it, I have added an invitation on the repo for you, it should enable changes if you accept it

If it does not, I will fork again on my personal account, sorry for the inconvenience.

src/metadata/metadata-storage.ts
```diff
@@ -286,4 +310,16 @@ export class MetadataStorage {
     }
     return authorizedField.roles;
   }
+
+  private findClassExtensions(target: Function): ExtensionsMetadata {
```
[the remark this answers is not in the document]
hihuz: Yeah this was also one of my initial ideas, but I couldn't find a way to make it right to the compiler, e.g.:
```
extensions.filter(
        entry => entry.target === target && (!(entry.fieldName || entry.fieldName === fieldName),
)
```
is not correct because `entry` can be of a type `ExtensionsClassMetadata` (which doesn't have `fieldName` property).

So, the best way I can think of to abstract this for now is this: b19c4afede1dbc7c5a8de0eda240ee89cc2b4d39 but I'm not really happy with the `any` cast.

If you have a better idea please correct it 🙂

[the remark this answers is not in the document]
MichalLytek: I know, union of array is tricky in TS. Changed it 😉

examples/extensions/authorizer.middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    if (roles.length > 0 && !user.roles.some(role => roles.includes(role))) {
+      // if the roles don't overlap, restrict access
+      throw new UnauthorizedError();
+    }
+  }
+
+  // grant access in other cases
+  await next();
```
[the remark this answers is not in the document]
hihuz: True, I have updated it in this commit: 363f8d3ab92e29c9960fd14bc017e51e79e30116

examples/extensions/authorizer.middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  return {
+    ...parentAuthorizationExtensions,
+    ...returnTypeAuthorizationExtensions,
+    ...fieldAuthorizationExtensions,
+  };
+};
+
+export const AuthorizerMiddleware: MiddlewareFn = async (
+  { context: { user }, info }: { context: Context; info: GraphQLResolveInfo },
```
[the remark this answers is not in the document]
hihuz: Indeed, updated here: 45a43d3393c91f34b2ac875ea2fd598bd77ea8f2

examples/extensions/authorizer.middleware.ts
hihuz: I tried to do this in this commit: cf7aa50de2207adff88a28b0f54b4411b34b3647
it seems a bit more complicated to me as we are not really interested in the whole config here and just the extensions though 🤔

examples/extensions/authorizer.middleware.ts
hihuz: I had already added the logger middleware in the examples as well 🙂 `examples/extensions/logger.middleware.ts`

My idea was to show we can define custom decorators on top of @Extensions with one middleware, but also that we can use @Extensions directly with another middleware. But I can also combine both approaches with one middleware 🙂I attempted to do it in this commit: 6617818ad896965a51fa55500d5a0ca2147112f8, I removed the whole authorization thing and only kept and extended the logger example. 

What do you think? I can also revert it if the previous iteration looked better

tests/functional/extensions.ts
```diff
@@ -0,0 +1,268 @@
...
+        const { mutationWithMultipleExtensionsDecorators } = schema.getMutationType()!.getFields();
+        expect(mutationWithMultipleExtensionsDecorators.extensions).toEqual({
+          first: "first mutation value",
+          second: "second mutation value",
+          third: "third mutation value",
+        });
+      });
+    });
+
+    describe("ObjectType", () => {
```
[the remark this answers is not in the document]
hihuz: Hmm, I am testing `@ObjectType` fields in this describe already: https://github.com/MichalLytek/type-graphql/pull/521/files#diff-f89e323ec29c3cd58ea94a7b10d388afR158

Or maybe you meant something else? Let me know if you have something more specific in mind, for now I'm not sure what I should add or change there

examples/extensions/logger.middleware.ts
```diff
@@ -0,0 +1,55 @@
...
+const extractParentConfig = (info: GraphQLResolveInfo): GraphQLObjectTypeConfig<any, any> =>
+  info.parentType.toConfig();
+
+const extractLoggerExtensionsFromConfig = (
+  config: GraphQLObjectTypeConfig<any, any> | GraphQLFieldConfig<any, any>,
+) => (config.extensions && config.extensions.log) || {};
+
+const getLoggerExtensions = (info: GraphQLResolveInfo) => {
+  const fieldConfig = extractFieldConfig(info);
+  const fieldLoggernExtensions = extractLoggerExtensionsFromConfig(fieldConfig);
```
MichalLytek: `Loggern` - typo

examples/extensions/logger.middleware.ts
```diff
@@ -0,0 +1,55 @@
+import { Service } from "typedi";
+import { GraphQLResolveInfo, GraphQLFieldConfig, GraphQLObjectTypeConfig } from "graphql";
+
+import { MiddlewareInterface, NextFn, ResolverData } from "../../src";
+
+import { Context } from "./context.interface";
+import { Logger } from "./logger.service";
+
+const extractFieldConfig = (info: GraphQLResolveInfo): GraphQLFieldConfig<any, any> => {
```
MichalLytek: Move `extractFieldConfig` and `extractParentConfig` into helpers - they are a generic `info` parse helpers

examples/extensions/resolver.ts
```diff
@@ -0,0 +1,42 @@
...
+      title,
+      description,
+      ratings: [],
+    });
+    this.recipesData.push(newRecipe);
+    return newRecipe;
+  }
+
+  @Logger("This message will not be logged")
+  @Logger("It will be overridden by this one")
```
MichalLytek: We have the rules described in docs - I think it's enough to show a real use case of the feature to demonstrate when it's helpful

tests/functional/extensions.ts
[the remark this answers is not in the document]
MichalLytek: That's ok, it was a bit missleading that input type fields were scoped in a different place with a `should add extensions to input type fields` test 😉

examples/extensions/authorizer.middleware.ts
MichalLytek: Yes, but for other extensions usage we might reuse that helpers, that's why it's better to have them extracted. It's also easier to read than the 20 lines monolith.

examples/extensions/authorizer.middleware.ts
MichalLytek: Looks ok I think 😉

examples/extensions/logger.middleware.ts
hihuz: 🤦‍♂ sorry for this, fixed here: 546fadb4de3d51b26f0d55599cac42cca6ea2a81

examples/extensions/resolver.ts
hihuz: Do you mean that this specific example of adding an extension twice on the same field should be remove and that showing it in docs is enough? If so, I have removed this example here: 531cc23820cf91f4c0ad26fcadb23afe9cf31ffd

Hope I got it properly

examples/extensions/logger.middleware.ts
hihuz: You mean to extract them to `src/helpers` correct? If so, I have done it here: 1779546e84b9ba1085ea81503fb280f781d87b7c

I have also added some basic tests for them as I think it's more proper if they are part of the actual source code and not just examples.

I have also renamed `extractParentConfig` -> `extractParentTypeConfig`, seemed more appropriate when using it, but I can change it back of course if you prefer.

MichalLytek: > You mean to extract them to src/helpers correct?

No, into example helpers, to not pollute the log middleware code

hihuz: Ah, alright my mistake, have relocated them into example helpers in this commit then: 19b68eece62205dc6e97b69d4bebc898a6e7a3ad

MichalLytek:
```
TypeError: Cannot read property 'log' of undefined",
  at LoggerMiddleware.use 
```
@hihuz I think you've forgot to setup container in `buildSchema` and haven't run the example 😄

hihuz:
@MichalLytek Thank you very much for all the time you spent looking at the PR and for all the feedback, much appreciated 🙂

Apologies for the remaining issue with the examples 🤦‍♂ and thank you for fixing them 👍

PR #537, opened 2020-02-12
fix: nullable=true behavior w/ arrays

dobrynin:
Hi. I noticed that `{ nullable: true}` was not behaving as described in the documentation. I believe this change fixes it. @MichalLytek can you take a look? Thank you!


dobrynin:
Hm, on further inspection, it looks like maybe graphql-voyager was incorrectly displaying the schema and that type-graphql was actually generating the schema correctly. I believe what I was seeing was actually caused by: https://github.com/APIs-guru/graphql-voyager/issues/151

I'll close this.

PR #538, opened 2020-02-13
Find scalars by name in convertTypeIfScalar

tristanMatthias:
## Problem
I am building a [CLI](https://github.com/tristanMatthias/brix/) that uses this amazing repo (Thanks! 🎉), but I've been having an issue with:

https://github.com/MichalLytek/type-graphql/blob/master/src/helpers/types.ts#L18

This issue is when a global package uses input scalars from the same package, but one is installed locally and the other is global, the check fails and I get a nasty error:

```
UnhandledPromiseRejectionWarning: Error: Cannot determine GraphQL input type for data
```

## Solution
This very tiny solution allows for an additional lookup by the name of the scalar itself. That way, no matter what version, or where it's loaded from, it will still resolve.

### DIFF
--- a/src/helpers/types.ts
+++ b/src/helpers/types.ts
@@ -18,7 +18,9 @@ export function convertTypeIfScalar(type: any): GraphQLScalarType | undefined {
   if (type instanceof GraphQLScalarType) {
     return type;
   }
-  const scalarMap = BuildContext.scalarsMaps.find(it => it.type === type);
+  const scalarMap = BuildContext.scalarsMaps.find(
+    it => it.type === type || it.scalar.toString() === type.toString(),
+  );
   if (scalarMap) {
     return scalarMap.scalar;
   }


MichalLytek:
> This issue is when a global package uses input scalars from the same package, but one is installed locally and the other is global

You should always have only one version of `graphql` package installed in your `node_modules`.

For simple apps `npm dedupe` is enough, for splitted packages lerna with deps hoisting is good, and for lib authors peer dependencies are a correct way to manage that.

tristanMatthias:
Yeah, I agree! But this is a slightly different case (unless I'm missing something), where one `graphql` is coming from the CLI globally installed, and the other is coming from the local `graphql`. Unfortunately in this case lerna/`npm dedupe` can't address this (unless there is a way for local imports to resolve to global modules, but that doesn't feel like a good approach).

Let me know if I'm missing something here. It's a bit of an obscure use case (the server running from the CLI). Either way, love this project a lot, I use it for everything I do.

MichalLytek:
> where one graphql is coming from the CLI globally installed, and the other is coming from the local graphql. 

What does you CLI is doing? Running a local code? 

Describe me your use case more, instead of providing a potential solution 😉

tristanMatthias:
Ok sure thing sorry!

I have a CLI (`@brix/cli`) that runs an instance of `type-graphql` and `apollo-server-express`. 
Now, as a feature of what I'm trying to build, I want to have as little configuration as possible for the user. Essentially just defining models and entities with `type-graphql`. The CLI then reads over the directory specified in a config file, and loads all the entities in that directory directly into the CLI's instance of `type-graphql`.

What this means is that the user will import a library like `graphq-type-json`, which then causes a bug because the `instanceof GraphQLScalarType` and the `===` for the `convertTypeIfScalar` function always fails. 

My potential solution (and sorry if it was too forward) is to also allow for an additional check based on the name of the scalar, not just the type comparison. 

Hope that makes sense, let me know if I should clarify anything further. Thanks for your patience @MichalLytek !

MichalLytek:
> where one graphql is coming from the CLI globally installed, and the other is coming from the local graphql

> The CLI then reads over the directory specified in a config file, and loads all the entities in that directory directly into the CLI's instance of type-graphql.

Sounds like a source of trouble. What will happen if the global CLI deps are not matching? User's code is using new TypeGraphQL v0.18 and your embedded version in CLI can't process it correctly. What if the new TypeGraphQL version uses some new GraphQL feature like extensions from v14.5 that your CLI might not use yet? I think will crash 😱 

So if you run the user's local code, you have to install the CLI locally, just like Create React App is doing with `react-scripts`.

tristanMatthias:
Ah I was afraid that was going to be the case. Ok thanks @MichalLytek

PR #539, opened 2020-02-13
refactor: more helpful error message

tomanagle:
This PR includes a small change to an error message to make it more informative and help the developer fix the issue.

All tests have been run and pass.

### DIFF
--- a/src/schema/schema-generator.ts
+++ b/src/schema/schema-generator.ts
@@ -658,7 +658,9 @@ export abstract class SchemaGenerator {
       }
     }
     if (!gqlType) {
-      throw new Error(`Cannot determine GraphQL input type for ${typeOwnerName}`!);
+      throw new Error(
+        `Cannot determine GraphQL input type for ${typeOwnerName}. Make sure you have used the correct directive.`!,
+      );
     }
 
     const { nullableByDefault } = BuildContext;


src/schema/schema-generator.ts
```diff
@@ -658,7 +658,9 @@ export abstract class SchemaGenerator {
       }
     }
     if (!gqlType) {
-      throw new Error(`Cannot determine GraphQL input type for ${typeOwnerName}`!);
+      throw new Error(
+        `Cannot determine GraphQL input type for ${typeOwnerName}. Make sure you have used the correct directive.`!,
```
MichalLytek: Wait, what? Directives? How they affect input type search? 😕

tomanagle: If you don't put a directive on the class and reference it in another class with @ObjectType or @InputType, you get this error but it doesn't tell you how to fix it.

MichalLytek: This are directives:
https://typegraphql.ml/docs/next/directives.html

`@ObjectType` or `@InputType` are decorators, so I think your new error message would not help anyone except you 😄

tomanagle: Right, no need to be sarcastic with people trying to help with poor, unhelpful error messages.

MichalLytek: You've even mislead me with this directives 😉 

The problem is that the "Cannot determine type" does not mean that you need to use a decorator. Somebody might put in `type =>` parameter an enum without calling `registerEnumType`, an object type class instead of input type, etc.
So it's not so trivial to add a proper and useful error message in that case 😕

PR #541, opened 2020-02-16
"aggregate" is now "count"

FluorescentHallucinogen:


MichalLytek:
Thanks @FluorescentHallucinogen but as you can see, neither `aggregate` nor `count` is now supported because it wasn't fully implemented in Prisma:
https://github.com/MichalLytek/type-graphql/blob/98b10664289df024a1060a014f67f7c5a9ba409c/src/generator/config.ts#L14-L19

I will update the name in docs when I update the corresponding code while adding support to the newest Prisma 2 preview release 😉

MichalLytek:
@FluorescentHallucinogen What is your source of the information about this change? I can't see `preview022` nor any changes in DMMF 😕

PR #567, opened 2020-03-07
Fix interface resolvers with custom schemaName

### LINKED ISSUE #566: Field name argument doesn't work with interface types
**Describe the bug**
```ts
@InterfaceType
abstract class Animal {
  @Field(() => String, { name: "animalName" })
  abstract getAnimalName: () => Promise<string>;
}

@ObjectType({ implements: [Animal] })
export class Dog {
  async getAnimalName() { return "Fido"; }
  /* ... */
}
```

The generated schema is along the lines of
```graphql
interface Animal {
  animalName: String!
}
```
but if you try to query, you get a GraphQL error along the lines of
```
Error: Cannot return null for non-nullable field Dog.animalName
```
seemingly because the mapping from `animalName` to `getAnimalName` hasn't happened.

**Addendum:**
I see that `name` isn't supported in some other places (#263) but this doesn't seem like the same situation.

travigd:
Fixes #566 

New to this codebase, open to feedback!

~~As an aside, the [`defaultFieldResolver` function](https://github.com/graphql/graphql-js/blob/fe878eef98aba23df988e858dbafa029d6351bd1/src/execution/execute.js#L1187) just does the _check if this is a function, if so, call it, otherwise return it as is_ thing.~~

**Edit:** I changed to use the `createBasicFieldResolver` function. It looks like it could also be extended to the advanced resolver type to support external interface resolvers? Also, I think that the status quo (before this PR) wouldn't apply any auth checking to fields defined on interfaces since it doesn't run any middlewares.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - fix transforming and validating nested inputs and arrays (#462)
 - remove duplicated entries for resolver classes that use inheritance (#499)
 - **Breaking Change**: stop returning null for `GraphQLTimestamp` and `GraphQLISODateTime` scalars when returned value is not a `Date` instance - now it throws explicit error instead
+- fix using `name` option on interface fields (#567)
 ### Others
 - **Breaking Change**: change build config to ES2018 - drop support for Node.js < 10.3
 - **Breaking Change**: remove deprecated `DepreciationOptions` interface
--- a/src/schema/schema-generator.ts
+++ b/src/schema/schema-generator.ts
@@ -16,7 +16,6 @@ import {
   GraphQLUnionType,
   GraphQLTypeResolver,
   GraphQLDirective,
-  defaultFieldResolver,
 } from "graphql";
 import { withFilter, ResolverFn } from "graphql-subscriptions";
 
--- a/src/schema/schema-generator.ts
+++ b/src/schema/schema-generator.ts
@@ -235,15 +235,7 @@ export abstract class SchemaGenerator {
                   fieldsMap[field.schemaName] = {
                     description: field.description,
                     type: this.getGraphQLOutputType(field.name, field.getType(), field.typeOptions),
-                    resolve: (source, args, contextValue, info) => {
-                      if (field.name !== field.schemaName) {
-                        return defaultFieldResolver(source, args, contextValue, {
-                          ...info,
-                          fieldName: field.name,
-                        });
-                      }
-                      return defaultFieldResolver(source, args, contextValue, info);
-                    },
+                    resolve: createBasicFieldResolver(field),
                   };
                   return fieldsMap;
                 },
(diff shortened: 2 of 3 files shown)


travigd:
As an aside, I tried to create a new test case for this but kept having weird errors.

I tried to create a new interface:
```ts
      @InterfaceType()
      abstract class FooInterface {
        @Field()
        fooInterfaceField: string;
      }
      @ObjectType({ implements: FooInterface })
      class FooImplementation implements FooInterface {
        fooInterfaceField: string;
        @Field()
        otherField: string;
      }
```

but whenever I wrote a query that returned `FooInterface`, I got a GraphQL error:
```
GraphQLError: Runtime Object type "FooImplementation" is not a possible type for "FooInterface".
```

Which was... strange.

travigd:
> As an aside, I tried to create a new test case for this but kept having weird errors.

Figured out that this was because of #568. I can re-write the test case to not latch onto the existing interfaces/objects.

@MichalLytek Assuming this looks good, would it be possible to backport this and release 0.17.6? I'm having some issues in my project that would be solved with this -- as well as the existing issue where middlewares are not applied for interface fields being a potential(?) security issue.

(and thanks as always for the awesome project!!!)

MichalLytek:
I feel like it's a treatment of symptoms, not the cure for the disease.

The problem is that we copy the interface field config, not the metadata, so as interfaces has no resolvers, object types does not have that too:
https://github.com/MichalLytek/type-graphql/blob/1ed59021c1df21aee484f25f83d509d2833ff464/src/schema/schema-generator.ts#L352-L363

Implementing `resolve` accidentally solves that problem because it's now copied and applied to object type.

So I would suggest changing that part to fix the field name mapping issue 😉

travigd:
I'm not sure why that's an issue. Or, well, I'm not really sure what the alternative is.

Since we are literally copying the fields from the interface to the object, and the interface is the one that knows how to resolve the field, why not just implement `resolve` there?

https://github.com/MichalLytek/type-graphql/blob/1ed59021c1df21aee484f25f83d509d2833ff464/src/schema/schema-generator.ts#L225-L241

Here, when we're generating `interfaceTypesInfo`, we lose all information about the distinction between the property name and the `schemaName`.

I can think of one alternative that **doesn't** do this, but I'm not sure if it's any better:

Forgo copying the fields and "manually" construct them from the original interface metadata. This would mean in `ObjectType`, we first generate an array of `FieldMetadata` and then once we've compiled the array, _then_ go create a bunch of `GraphQLField`s based on that array. This allows us to keep the information about property name vs `schemaName`.

MichalLytek:
It makes no sense to change this logic as it might affect some other parts and result in regression. 

Let's fix that in this simple way, as `vNext` gonna have this inheriting part solved in a different way anyway 😉

travigd:
@MichalLytek Can this particular change be backported to the 0.17.x line? I'm running into an issue where an interface field isn't running the middleware because of this.

MichalLytek:
I'm afraid I don't have time for this 😞 
Please use your fork if you're afraid to use `0.18-beta` which gonna be promoted to `1.0.0-rc` soon 😉

PR #595, opened 2020-04-04
Automatically register object types implementing used interface type

### LINKED ISSUE #568: Orphaned Types
I've been opening up several issues on this project lately (because I think it's pretty darn cool!). I'd like to devote this one to talking about orphaned types as it doesn't seem there's a single place where that discussion has been taking place.

 In the 0.18 beta, there is a concept of _orphan types_, that is, types that exist (are declared with decorators) but aren't explicitly referenced in the schema. The reason for this is the desire to support non-global schemas (e.g., we want to support more than two schemas in the same node process).

The main issue with the current state of things (as of the 0.18-beta) is that it doesn't work super well for interfaces (at the very least, the behavior is unintuitive). Suppose you have an interface `Animal`, a query `getAnimal` that returns an `Animal`, and several implementations such as `Cat`, `Dog`, etc.. Currently, if `Cat` and `Dog` aren't included elsewhere in the schema (e.g., in a query that explicitly returns a `Dog` instead of an `Animal`), then `Cat` and `Dog` are considered oprhaned types and are not added to the GraphQL schema by default. The end result of this is that there is no way to actually use the `Animal` type (since all implementations aren't actually defined in the schema).

### Explicit `orphanedTypes` argument (status quo)
This is the status quo.
* **Pro:** Avoids all potential issues with duplicate schemas
* **Con:** Has issues with interfaces (described above)
* **Con:** All orphaned types have to be supplied in `buildSchema` which makes it harder to be modular
* **Con:** This is very unintuitive behavior (especially if you're used to the existing type-graphql behavior)

### Add an `implementedBy` field to `@InterfaceType` decorator
Proposed in https://github.com/MichalLytek/type-graphql/issues/110#issuecomment-562233797.
* **Pro:** More modular than first option
* **Con:** Likely to have issues with circular dependencies

### Auto-register interface implementations
This is my preference. Essentially, internally have some metadata associated with each interface type, and when the interface type is added to the schema, also register all known implementations at that time.
* **Pro:** Fully modular
* **Pro:** No circular dependencies
* **Pro:** More intuitive (at least with respect to current type-graphql usage)
* **Con:** It's possible to "forget" to include a class (if it's not explicitly imported). This is the case with 0.17 and also not that big of a deal since if it hasn't been imported, it can't be instantiated anyway (assuming you're using class instances).

### Explicit Schema Registration
This is a huge departure from the way things are now, but I thought I'd mention it. In (for example) Python's Flask, you can register routes using decorators:
```
@app.get("/foo")
def get_foo():
    ...
```

We could do something similar where we have an explicit schema object and the decorators are instance methods.
```ts
@schema.ObjectType()
class ... { /* ... */ }
```

Alternatively, the schema could be an option to the decorator, and if not provided, use a global default.
```ts
@ObjectType({
  // If omitted, this defaults to the "global" schema
  schema: internalSchema
})
class ... { /* ... */ }
```

* **Pro:** Very explicit about what objects are associated with which schemas
* **Con:** Huge departure from the way things are today
* **Con:** Circular dependency issues (Flask has _lots_ of these)

---

The third proposal (_Auto-register interface implementations_) is my favorite I'd be happy to take a stab at this. It's also a minor change from the way things are in the 0.18-beta codebase.

MichalLytek:
This PR adds a feature that automatically emit in schema all the object types that implements the interface types referenced in schema definition.

It also adds an option to disable the auto register feature per interface type, like the `Node` interface case for Relay-based system, in case of multiple schemas isolation.

Closes #568

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - optimize resolvers execution paths to speed up a lot basic scenarios (#488)
 - add `@Extensions` decorator for putting metadata into GraphQL types config (#521)
 - add support for defining arguments and implementing resolvers for interface types fields (#579)
+- add `{ autoRegisterImplementations: false }` option to prevent automatic emitting in schema all the object types that implements used interface type (#595)
 ### Fixes
 - refactor union types function syntax handling to prevent possible errors with circular refs
 - fix transforming and validating nested inputs and arrays (#462)
--- a/docs/bootstrap.md
+++ b/docs/bootstrap.md
@@ -9,7 +9,7 @@ After creating our resolvers, type classes, and other business-related code, we 
 To create an executable schema from type and resolver definitions, we need to use the `buildSchema` function.
 It takes a configuration object as a parameter and returns a promise of a `GraphQLSchema` object.
 
-In the configuration object you must provide a `resolvers` property, which can be an array of resolver classes:
+In the configuration object we must provide a `resolvers` property, which can be an array of resolver classes:
 
 ```typescript
 import { FirstResolver, SecondResolver } from "../app/src/resolvers";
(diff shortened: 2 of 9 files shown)


travigd:
Thanks for this, it looks good to me. :^)

PR #611, opened 2020-04-20
docs(getting-started): fix repeated name

dandv:


MichalLytek:
Thanks @dandv ❤️ 

It was fixed in readme but not in docs:
image

PR #626, opened 2020-05-06
fix(schema): Fix directives cannot be extends

lijianzhang:
For Example:

```typescript
class INode {
    @Field(() => ID)
    id!: string

    @Directive('@formatDate')
    @Field(() => Timestamp)
    createdAt: Date;

    @Directive('@formatDate')
    @Field(() => Timestamp)
    updatedAt: Date;
}

//  People class cannot extends INode Filed directives  
@ObjectType()
class People extends INode {

}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 <!-- ## Unreleased -->
 <!-- here goes all the unreleased changes descriptions -->
+- fix `@Directive()` cannot be extends by subclasses 
 
 ## v1.0.0-rc.1
 ### Features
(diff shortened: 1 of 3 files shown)


lijianzhang:
@MichalLytek Thanks for your reminder. I have added test cases 😅

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@
 - fix calling field resolver without providing resolver class to `buildSchema`
 - fix generated TS union type for union type of object type classes extending themselves (#587)
 - fix using shared union and interface types in multiple schemas when `resolveType` is used
+- fix `@Directive()` cannot be extends by subclasses
```
MichalLytek, line 34: This entry it's for the "unreleased" section as merging this PR won't overwrite `1.0.0-rc.1` 😉

tests/functional/directives.ts
```diff
@@ -378,6 +397,18 @@ describe("Directives", () => {
       });
     });
 
+    describe("SubInputType", () => {
```
MichalLytek, line 400: it shouldn't be own describe block with type name, but rather just a `it` test case like "adds inherited field directives to input type fields while extending input type class"

tests/functional/directives.ts
```diff
@@ -416,6 +447,45 @@ describe("Directives", () => {
         });
       });
     });
+
+    describe("SubObjectType", () => {
```
MichalLytek, line 451: same title issue

lijianzhang:
@MichalLytek  I have dealt with it. thank you for your patience to point out these problems

and `type-graphql` is awesome.

PR #635, opened 2020-05-25
Add info about shim for cypress

webnoob:
I've had to do the same browser shim in cypress because I am using a module which exports the type-graphql decorators. This PR adds that info to the docs.

### DIFF
--- a/docs/browser-usage.md
+++ b/docs/browser-usage.md
@@ -6,7 +6,7 @@ title: Browser usage
 
 Sometimes we might want to use the classes we've created and annotated with TypeGraphQL decorators, in our client app that works in the browser. For example, reusing the args or input classes with `class-validator` decorators or the object type classes with some helpful custom methods.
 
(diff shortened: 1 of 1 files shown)


MichalLytek:
TBH, I'm not sure about that.
The snippet looks almost exactly the same as the webpack one:
image

The only difference is the webpack preprocessor syntax:
https://github.com/cypress-io/cypress-webpack-preprocessor
which is Cypress specific and doesn't need to be explicitly features here.

If you really find that info missing and valuable, maybe just add a smaller section, not an h2, with just a link how to use the webpack config snippets with webpack preprocessor and tsloader for Cypress.

webnoob:
From my perspective, I don't really mind if it's there or not. I was just adding this as a heads up for users doing this as the error from cypress is about `utils1_promisify` is not a function. I had no clue that this was the issue and it was only a guess that this might be the cause.

I understand the change is the same but that's what's required to make type-graphql work with cypress. If you don't want that info included - no sweat - mine is working :)

MichalLytek:
@webnoob I've replaced the full guide with some simpler info which should be enough to know how to solve the problem with Cypress and browser shim. 

Let me know what do you think so I can merge this one.

webnoob:
Much more concise - perfect. Looks good and thanks for doing that (I should have done it that way to begin with really .. 🙄)

Thanks for all your hard work on behalf of us.

PR #639, opened 2020-05-28
Update graphql-query-complexity to ^0.6.0

rh389:
Currently there's a warning on installation alongside `graphql@15`:

`type-graphql > graphql-query-complexity@0.5.0" has incorrect peer dependency "graphql@^0.13.0 || ^14.0.0"`

`graphql-query-complexity@0.6.0` [recently widened its peer dependencies](https://github.com/slicknode/graphql-query-complexity/pull/30) to include `graphql@^15.0.0`. 

It dropped `legacyEstimator` and `fieldConfigEstimator` in the process, but it looks like this library doesn't use the former and has just moved away from the latter (both were already deprecated in `0.5.0`).

This PR updates the dependency to `^0.6.0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 <!-- here goes all the unreleased changes descriptions -->
 ### Features
 - **Breaking Change**: remove legacy array inference - now explicit array syntax (`[Item]`) is required
+### Fixes
+- Update `graphql-query-complexity` dependency to `^0.6.0` to fix peer dependency warning with `graphql@^15.0.0`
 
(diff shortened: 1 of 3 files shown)


MichalLytek:
@rh389 Thanks for the PR!

However, I've just updated all the deps, which superseeds your PR 🤐 
Closing via 38334d0 🔒

rh389:
“Duplicate” is a little harsh, but otherwise no problem! 😉

Cheers for the library 👍

MichalLytek:
Sorry, I don't have the "outdated" label and I found that less offensive than "invalid" 😄

PR #640, opened 2020-05-28
missing 'get' for simple field resolver

madmaxlax:


MichalLytek:
@madmaxlax Thanks for reporting!

However, getter can't have any arguments, that's why `averageRating` is defined as the class method. So your fix is unfortunately invalid 😞

madmaxlax:
> @madmaxlax Thanks for reporting!
> 
> However, getter can't have any arguments, that's why `averageRating` is defined as the class method. So your fix is unfortunately invalid 😞

ah thank you for the insight. 
for some issues I was having with a simpler field resolver, adding `get`  helped but it took a while to figure out. and thought maybe the documentation example needed it too. 
i wasn't aware of the no arguments rule though

MichalLytek:
> i wasn't aware of the no arguments rule though

It is the standard TypeScript and JavaScript requirement, as getter = `object.property` access, not `object.propert(param)` like a method.

PR #642, opened 2020-06-04
docs: fix logic of complexity error statement

impulse:
I think it would make more sense to remove the `=` so it doesn't throw at 20 with "it is over 20"


MichalLytek:
Thanks for spotting this! 💪 

Can I ask you to update also the previous versions docs too?
You can find them in `/website/versioned_docs` folders.

Thanks! 🙏

impulse:
@MichalLytek done, I also adjusted it for the example file/

Thanks for the project, really enjoying it so far!

Issue #117, opened 2018-07-21
Integration with typegoose


danpaugo:
Hey @19majkel94 I am trying to use typegoose with type-graphql. Following are the relevant code snippets from my user.model.ts, user.service.ts and user.resolver.ts files.
user.model.ts
```
@ObjectType()
@pre<User>('save', async function () {
  var user = this;
  if (user.isModified('password')) {
    user.password = await bcrypt.hash(user.password, SALT_WORK_FACTOR)
  }
})
export class User extends Typegoose {
  @Field(() => ID)
  _id: string;
  
  @Field(() => Date)
  createdAt: Date;

  @Field(() => Date)
  updatedAt: Date;

  @prop({ required: true })
  @Field(() => String, { nullable: false })
  firstName: string;


  @prop({ required: true })
  @Field(() => String, { nullable: false })
  lastName: string;

  @Field(() => String, { nullable: false })
  @prop({ required: true })
  email: string;

  @prop({ required: true })
  password: string;
}

export const UserModel = new User().getModelForClass(User, { schemaOptions: { timestamps: true } });
```

user.service.ts
```
import { UserModel, User } from '../models/user.model';

export const getUserById = async (userId: string): Promise<User> => {
  const user: User = await UserModel.findOne({ _id: userId });
  if (!user) {
    throw new UserInputError(`User with userId ${userId} not found`);
  }
  user.password = null;
  delete user.password;
  return user;
};
```

user.resolver.ts
```
import { User } from '../models/user.model';
import { getUserById } from '../services/user.service';

@Resolver(() => User)
export class UserResolver {
  @Query(() => User, { nullable: false })
  @UseMiddleware(Auth)
  async user(@Arg('_id') _id: string, @Ctx() context: Context): Promise<User> {
    return getUserById( _id);
  }
 }
```

Whenever I try to fire the user resolver I get the following error
```
{
  "data": null,
  "errors": [
    {
      "message": "Expected value of type \"User\" but got: { _id: 5b530bec3b461c03cef31a53,\n  firstName: 'Danny',\n  lastName: 'Paul',\n  email: 'danny@paul.io',\n  password: null,\n  createdAt: 2018-07-21T10:33:26.763Z,\n  updatedAt: 2018-07-21T10:33:26.763Z,\n  __v: 0 }.",
      "locations": [
        {
          "line": 9,
          "column": 3
        }
      ],
      "path": [
        "user"
      ],
      "extensions": {
[...]

PR #657, opened 2020-06-18
Changes arrayProp to prop as arrayProp is deprecated

smolinari:
Hope this is ok. 

Refers to this: https://github.com/MichalLytek/type-graphql/issues/117#issuecomment-645810830

Scott

### DIFF
--- a/examples/typegoose/entities/recipe.ts
+++ b/examples/typegoose/entities/recipe.ts
@@ -23,7 +23,7 @@ export class Recipe {
   description?: string;
 
   @Field(type => [Rate])
-  @Property({ items: Rate, default: [] })
+  @Property({ type: () => Rate, default: [] })
   ratings: Rate[];
 
   @Field(type => User)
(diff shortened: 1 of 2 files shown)


smolinari:
@MichalLytek - Sorry, but I'm not sure what your comment is implying. 

Scott

MichalLytek:
` @Property({ items: Rate, default: [] })`

smolinari:
> ` @Property({ items: Rate, default: [] })`

Yes, `Rate` is a Class and a sub-document, right? 

Scott

MichalLytek:
@smolinari `items` is deprecated, use `type` from 7.2.0

smolinari:
Ah. Geez. Thanks for catching that. 

Scott

smolinari:
Hope that is correct now. Docs aren't super clear on what the value for `type` should be. I gathered this from the Typegoose change discussions.

Scott

PR #660, opened 2020-06-20
Update scalars.md specify which is the default value for dates

brunolm:


MichalLytek:
Line 69:
> By default, TypeGraphQL uses the ISO date format

Do you really find that explicit name so useful, even when typescript autocompletes one of the two string values?

brunolm:
@MichalLytek Oh you're right, the information is already there. I didn't see that when I looked over quickly. I also had trouble to find the default value for `nullable`.

The information is there but I guess to find it I have to search text withing the page.

Maybe I would find this information easier if it was near the first usage example

```
@Field({ nullable: true }) // default value is `false`
  averageRating?: number;
```

```
dateScalarMode: "timestamp", // "isoDate" (default) or "timestamp"
```

PR #669, opened 2020-07-08
Add Next.js browser shim and update headers/format

desmap:
[Updated layout since the PR before was not consistent] Since Next.js is quite popular and it took me some time to figure this out, this might be helpful for others

### DIFF
--- a/docs/browser-usage.md
+++ b/docs/browser-usage.md
@@ -6,7 +6,14 @@ title: Browser usage
 
 Sometimes we might want to use the classes we've created and annotated with TypeGraphQL decorators, in our client app that works in the browser. For example, reusing the args or input classes with `class-validator` decorators or the object type classes with some helpful custom methods.
 
-Since TypeGraphQL is a Node.js framework, it doesn't work in a browser environment, so we may quickly get an error, e.g. `ERROR in ./node_modules/fs.realpath/index.js` or `utils1_promisify is not a function`, while trying to build our app with Webpack. To correct this, we have to configure Webpack to use the decorator shim instead of the normal module. We simply add this plugin code to our webpack config:
+Since TypeGraphQL is a Node.js framework, it doesn't work in a browser environment, so we may quickly get an error, e.g. `ERROR in ./node_modules/fs.realpath/index.js` or `utils1_promisify is not a function`, while trying to build our app e.g. with Webpack. To correct this, we have to configure bundler or compiler to use the decorator shim instead of the normal module.
+
+The steps to accomplish this are different, depending on the framework, bundler or compiler we use.
+However, in all cases, using shim makes our bundle much lighter as we don't need to embed the whole TypeGraphQL library code in our app.
+
+## CRA and similar
+
+We simply add this plugin code to our webpack config:
 
 ```js
 module.exports = {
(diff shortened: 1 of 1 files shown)


docs/browser-usage.md
```diff
@@ -22,17 +24,38 @@ module.exports = {
...
+    "paths": {
+      "type-graphql": ["node_modules/type-graphql/dist/browser-shim.ts"]
     }
   }
 }
 ```
 
+Then, `npm install -D tsconfig-paths` and enable it with `NODE_OPTIONS="-r tsconfig-paths/register"` in your environment variables setup.
+
 Thanks to this, our bundle will be much lighter as we don't need to embed the whole TypeGraphQL library code in our app.
```
MichalLytek, line 61: And now this sentence is out of context thanks to the new sections titles 😝

docs/browser-usage.md
```diff
@@ -22,17 +24,38 @@ module.exports = {
...
+
+```json
+{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "type-graphql": ["node_modules/type-graphql/dist/browser-shim.ts"]
     }
   }
 }
```
MichalLytek, line 56: I think it's enough to reference the `tsconfig.json` from above as its 1:1 the same

docs/browser-usage.md
```diff
@@ -22,17 +24,38 @@ module.exports = {
...
     "baseUrl": ".",
     "paths": {
-      "type-graphql": ["./node_modules/type-graphql/dist/browser-shim.ts"]
+      "type-graphql": ["node_modules/type-graphql/dist/browser-shim.ts"]
+    }
+  }
+}
+```
+
+## Next.js and similar
```
MichalLytek, line 42: The section titles are only misleading and should be removed

docs/browser-usage.md
```diff
@@ -22,17 +24,38 @@ module.exports = {
...
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "type-graphql": ["node_modules/type-graphql/dist/browser-shim.ts"]
     }
   }
 }
 ```
 
+Then, `npm install -D tsconfig-paths` and enable it with `NODE_OPTIONS="-r tsconfig-paths/register"` in your environment variables setup.
```
MichalLytek, line 59: I think it's too detailed - the CRA section doesn't show how to install and configure the `craco` package

docs/browser-usage.md
desmap: Yes if we remove the headings again one `paths` config code block should be enough. It did it 2x because some people (like me) just ignore anything written about 'Angular`, so I totally missed that piece of info for quite some time.

docs/browser-usage.md
desmap: Tbh, I would even elaborate more on this.

Next is quite sophisticated on how it does everything and it is by far the most used framework (CRA and Angular are a fraction of Next). You should def keep this snippet, otherwise your user won't get Next to work:

- Next's webpack config automatically pulls any `tsconfig` `paths` redirection and apply it to its compile step without the need to any webpack plugin like usually (which is great)
- However, you need to apply this as well to the runtime which is done by `node` which is hidden and abstracted away under the `next` CLI
- Since you can't access the underlying `node` you need to pass `tsconfig-paths/register` via the env `NODE_OPTIONS`

Once you now above life is easy but to get there took me +2 days and again since Next is the major React platform this README should focus on Next. The rest is nice-to-have but tbh, nobody cares.

Re craco, you should def add that the CRA scripts do not work anymore and craco's replacement script need to be used. 

Let me know how you want to the shim doc and i can rewrite all.

desmap: Or what we could also do, just forget all the framework-specific crap and educate the user and tell him that he need redirects for compile-time and build-time and depending on his framwork his does this with webpack-plugins, ts paths config and/or the tsconfig-paths module

hyusetiawan: i am vetting out next.js and type-graphql combo. After scouring for docs + tutorials + videos, I have a relevant perspective on this. In my personal opinion, the more detailed the better? It would definitely increase the dev ux which arguably increases adoption. I tried nexus then type-graphql, nexus has weird typescript errors that are explained in different places which made the whole setup rather painful, I have since abandoned the nexus route because typegraphql docs got me farther.

MichalLytek (approved):

MichalLytek:
I've rebased this branch and applied all the resolutions from the discussion. It's now ready to merge 😉 

@desmap Thanks for your contribution! 🙌

Issue #673, opened 2020-07-27
The current browser shim feature is barely usable, so code sharing is impossible in real production environments, it needs to be changed


desmap:
The UX of current browser shim feature is currently hardly usable and might be significantly improved.

**_tl/dr the proposed solution: `type-graphql` just knows if it shall deliver the real module or the shim via env var which is set in the Docker/k8s orchestration per service such as `TYPE_GRAPHQL_SHIM: 1`_**

**A short wrap-up about the current UX:** The shim is used by importing `typegraphl/dist/browser-shim.js` instead of `type-graphl`. So, you need to be aware that depending on your context/package you have to import either the real module or the shim. And it must be always clear always both (1) `tsc` at build time and (2) `node` at runtime must be aware of this. The user has to take care that both in 1 and 2 this "redirection" is *really* happening.

**The Problem:** While former is possible in many ways, most of the solutions are hacky and create complications and eventually just don't work together in a monorepo whose packages in production are reflected as isolated Docker containers/k8s services. I don't want to list all the problems but just give a hint: We face so many issues, setting up monorepos is already no trivial thing (just think about file organization or TS' quite restricted + complicated project references features which shakes up your whole dir structure, TS paths feature can't be used always), then we have three competing package managers, one can't do monorepos, another one which future is unclear and doesn't support ESM (yarn) and another which does not have an ecosystem yet (pnpm) and finally we need to think, implement and remember n * m solutions for this redirection (n = packages, m = 2 = a solution at build and at runtime). These n * m implementations need to be also interoperable. Good luck or tbh, it's not possible.

**The Solution**: `type-graphql` just knows if it shall deliver the real module or the shim via env var which is set in the Docker/k8s orchestration per service such as `TYPE_GRAPHQL_SHIM: 1`, _then_ we wouldn't need any hacky redirection which go deep in the build system nor dynamic imports which can't be setup DRY without top-level awaits for which you need ESM support. This list goes on forever and makes the current feature implementation, sorry to say absolutely useless.

Maybe there are better solutions but I know that current one is not really working. The build process/system was taking 99% of my time the last weeks and only because I wanted to share some types and class validators. I don't think that code generation —I could just `cpx` the files on change into some other packages—shouldn't be the solution for real and proper code sharing.

MichalLytek:
> The Solution: type-graphql just knows if it shall deliver the real module or the shim via env var

How it gonna prevent into bundling the whole `graphql-js` library then?

[...]

PR #674, opened 2020-07-28
Enable shim via env var TYPE_GRAPHQL_SHIM=1

desmap:
Current system leads to many complications, compare also https://github.com/MichalLytek/type-graphql/issues/673


MichalLytek:
I found this obsolete after merging #669 🔒

PR #718, opened 2020-10-06
fix: Allow `readonly string[]` for `Subscription` topics

rh389:
I just ran into this while attempting to pass something like `['Topic1', 'Topic2'] as const` to subscription options - there's no reason to restrict to mutable arrays here.


MichalLytek (changes requested):
Please try to reproduce the error in tests, push the failing commit and then after CI fail push the fixing commit.

rh389:
Ach, looks like the issue starts upstream in `graphql-subscriptions`, (https://github.com/apollographql/graphql-subscriptions/pull/234) where the change would be breaking.

Thanks for the quick response anyway, closing.

PR #725, opened 2020-10-13
use optional chaining to get value inside reflectedType

techo207:
**Describe the Bug**
I'm not sure if it is a bug, or my specific use case.
I'm creating a custom decorator that inject a generic find function as  field function in the ObjectType class. to achive it i'm using the Args decorator as a function but it throws a error in runtine.

> Cannot read property '0' of undefined

**To Reproduce**
```ts
@ArgsType()
class Ca{
   @Field(() => String)
   a: string;
}

@ObjectType()
class A {
    @ConnectionField( () => ProductFindArgs )
    prueba;
}

export function ConnectionField( findArgs: any, options?: FieldOptions ) {
	return function ConnectionDecorator( target: any, propertyKey: string | symbol ) {
		const newPropertyName = `_${String( propertyKey )}Connection`;
		const field = Field( () => String , { name: String( propertyKey ) } );
		const args = Args(findArgs);
		target[newPropertyName] = async function ConnectionFieldResolver( { a }: any ) {
			return findMany( a );
		};
		const descriptor :PropertyDescriptor = { configurable: true, enumerable: true, writable: true };
		args( target, newPropertyName, 0 );
		field( target, newPropertyName, descriptor );
	};
}
```

**Expected Behavior**
I expected that it create  a function to resolve prueba field inside object type, but it throws a error

I can resolve the problem just adding a optional chaining in line 11 inside /dist/helpers/findType.js , just like 
```ts
 metadataDesignType = reflectedType?.[parameterIndex];

```

**Logs**
at .../node_modules/type-graphql/dist/helpers/findType.js:11
        metadataDesignType = reflectedType[parameterIndex];
                                          ^
TypeError: Cannot read property '0' of undefined
    at .../node_modules/type-graphql/dist/helpers/findType.js:11:43)
    at Object.getParamInfo (.../node_modules/type-graphql/dist/helpers/params.js:10:49)
    at. .../node_modules/type-graphql/dist/decorators/Args.js:12:25

- OS: mac os 
- Node 14.0
- Package version  lastest
- TypeScript version  4.0.3

### DIFF
--- a/src/helpers/findType.ts
+++ b/src/helpers/findType.ts
@@ -41,7 +41,7 @@ export function findType({
     propertyKey,
   );
   if (metadataKey === "design:paramtypes") {
-    metadataDesignType = (reflectedType as Function[])[parameterIndex!];
+    metadataDesignType = (reflectedType as Function[])?.[parameterIndex!];
   } else {
     metadataDesignType = reflectedType as Function | undefined;
   }


MichalLytek:
> I'm creating a custom decorator that inject a generic find function as field function in the ObjectType class. to achive it i'm using the Args decorator as a function but it throws a error in runtine.

Instead of fixing it from the wrong side, it's better to discuss your use case.
My question is: why you do such weird things instead of use the generic types feature?

Registering virtual fields dynamically is against TypeGraphQL goals as your args TS type has no notion about this property hence you can't use it in the resolver.

techo207:
I'm using the decorator to resolve a one to many relacion. A ObjectType have many Prueba and the function findMany inside the decorator resolve it with the where and order options inside the  ProductFindArgs inputType

MichalLytek:
> to resolve a one to many relacion

Please describe your use case in more detail, I still don't know how your printed graphql schema looks like, how the other classes looks like and why you need to call the decorator fn manually instead of using generic types and resolvers feature.

MichalLytek:
From what I understand, you have object type with simple field (let's say `bar` in English instead of your local `pueba`):

```ts
@ObjectType()
class Foo {
  @Field()
  bar!: string;
}
```

And what you want to achieve is that it has another `barConnection` field in graphql query can accept some args (`ProductFindArgs`) and will automatically resolve that relation using `findMany` function?

If so, why you can't just create a field resolver for that?
```ts
@ObjectType()
class Foo {
  bar!: FooBarObjectType[];

  @Field(returns => FooBarObjectType)
  barConnection(@Args() args: ProductFindArgs) {
    return findMany(args);
  }
}
```

Your snippet is missing the type of the field, which is why TS reflection fails, as well as your code.

Try to use less magic. The code will be more reliable and easier to understand by others.

MichalLytek:
Closing as superset by #733 🔒

PR #730, opened 2020-10-16
docs(interfaces): clarify chained interface syntax

jtassin:
Update the doc of a chained interfaces design.

With : 

```ts
@ObjectType({ implements: [IPerson] }) // it miss Node here
class Person extends IPerson {  
  @Field()
  name: string;
}
```

as a `Person` declaration the id type from `Node` will be missing in the Person type and the graphql schema will be wrong with the following error : 

```bash
Type Person must implement Node because it is implemented by IPerson.
Interface field IPerson.id expected but Person does not provide it.
```

Without this exemple, it took me a certain time to figure this error out.

PS: thx a lot for your lib :)

### DIFF
--- a/docs/interfaces.md
+++ b/docs/interfaces.md
@@ -71,12 +71,16 @@ class Node {
 }
 
 @InterfaceType({ implements: Node })
-class Person extends Node {
-  @Field()
-  name: string;
+class IPerson extends Node {
 
   @Field(type => Int)
   age: number;
+}
+
+@ObjectType({ implements: [Node, IPerson] })
+class Person extends IPerson {  
+  @Field()
+  name: string;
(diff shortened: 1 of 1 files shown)


MichalLytek:
@jtassin Your example without explanation doesn't give too much valuable info.

The error "Type Person must implement Node because it is implemented by IPerson" is clear and there's no any special thing to do than making `implements: [IPerson, Node]` which is straightforward.

The behavior is well described in GraphQL spec:

> Transitively implemented interfaces (interfaces implemented by the interface that is being implemented) must also be defined on an implementing type or interface.

MichalLytek:
Solved by d4bfe835 😉

PR #739, opened 2020-10-21
docs(subscriptions): fix arguments of the `subscribe` option

n1te1337:
This PR fixes the docs for the function arguments of the `subscribe` option (as per the `ResolverFn` [signature](https://github.com/apollographql/graphql-subscriptions/blob/master/src/with-filter.ts#L4)).


MichalLytek (approved):

MichalLytek (approved):
Thanks for spotting that issue! ❤️

PR #744, opened 2020-10-31
Add directives support to Interfaces

Superd22:
This PR adds support for using `@Directive` on InterfaceType (and its fields).

Directives are supported on interface as per [GraphQL specs](https://spec.graphql.org/June2018/#sec-Interfaces) and that support is widely implemented, see [Apollo](https://github.com/graphql/graphql-js/blob/a62eea88d5844a3bd9725c0f3c30950a78727f3e/src/language/directiveLocation.js#L22-L33) for example.

### Possible caveats
In this proposed implementation, any ObjectType implementing the interface (more precisely, whose typescript class extends the typescript class of the interface) will inherit its directives. This seems in line with the current behavior of TypeGraphQL.
Nevertheless, as it stands, users would **not** be able to __only__ add directive to the interface and not the objects implementing it.

I'd love insights about this behavior and if you think that's a problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 - **Breaking Change**: `AuthChecker` type is now "function or class" - update to `AuthCheckerFn` if the function form is needed in the code
 - support class-based auth checker, which allows for dependency injection
+- allow defining directives for interface types and theirs fields, with inheritance for object types fields (#744)
 
 ## v1.1.1
 ### Fixes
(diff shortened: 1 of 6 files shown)


MichalLytek:
Sorry @Superd22 for no response for this PR for over 2 months, I've totally forgot about this 😞 

> users would not be able to only add directive to the interface and not the objects implementing it

I think that's ok. If there will be some use cases for separating interfaces from object types, we can always add some global or scoped config for that 😉

MichalLytek (approved):
Looks great, thanks! ❤️

MichalLytek:
I've added changelog entry and updated docs 😉 Merging 🚀

PR #746, opened 2020-11-04
docs(subscriptions): Add how to integrate subscription servers without ApolloServer

acro5piano:
I think it is helpful to mention how to implement without `ApolloServer` for those who don't use it.

### DIFF
--- a/docs/subscriptions.md
+++ b/docs/subscriptions.md
@@ -194,6 +194,31 @@ const server = new ApolloServer({
 
 And it's done! We have a working GraphQL subscription server on `/subscriptions`, along with the normal HTTP GraphQL server.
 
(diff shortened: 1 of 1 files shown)


MichalLytek:
@acro5piano Thanks for the PR 😉 

However, I found it out of the scope of the TypeGraphQL docs.
You can use TypeGraphQL with any GraphQL server that supports `graphql-js`, thus it's not limited to only Apollo Server. It's mentioned in docs as it's the most popular and easy to use.

Putting more detailed instruction for less popular cases would bloat the whole docs which plenty of users found too stout and then ask for help on stuff described in docs.

So I'm rejecting this PR as I can't make even more harder to read by newcomers and because the subscriptions configuration is not related strictly to TypeGraphQL.

acro5piano:
@MichalLytek Okay! Thanks for checking it out.

PR #763, opened 2020-12-22
feat(types): use static fromModule from Target if exists

syadykin:
Hi

I faced the same issue as described in https://github.com/MichalLytek/type-graphql/issues/117 but proposed solution didn't work for me because my typegoose model have subschema in it:

```typescript
@ObjectType()
export class Avatar {
  @Field(() => Int)
  @prop({ required: true })
  public height!: number;

  @Field(() => Int)
  @prop({ required: true })
  public width!: number;

  @Field()
  public get url(): string {
    return "https://...";
  }
}

@ObjectType()
export class User {
  @Field(() => Avatar, { nullable: true })
  @prop({ _id: true, type: () => Avatar, required: false })
  public avatar?: Avatar;
}
```

When `type-graphql` serializes `User` model from mongoose document (which is returned by `find` static method) it looses all properties in `avatar` because of the way how mongoose document keeps data internally (they appeared available in `this._doc` which I also saw in other issues).

So there are couple of possible solutions how to avoid this:
1. use a kind of "standard" function shortcut for getting object data via `obj.toJSON()` or `obj.toObject()` but this could lead to unwanted results in already existing projects
2. use type class method, like `fromModel` which converts real data into type, which could be smth below for mongoose:

```typescript
@ObjectType()
export class Avatar {
  @Field(() => Int)
  @prop({ required: true })
  public height!: number;

  @Field(() => Int)
  @prop({ required: true })
  public width!: number;

  @Field()
  public get url(): string {
    return "https://...";
  }

  public static fromModel(data: DocumentType<Avatar>): Avatar {
    return Object.assign(new Avatar(), data.toObject());
  }
}

```

This solution also helps me with a more complex cases as well.

@MichalLytek do you think it worth merging?

### DIFF
--- a/src/helpers/types.ts
+++ b/src/helpers/types.ts
@@ -107,7 +107,9 @@ export function convertToType(Target: any, data?: object): object | undefined {
     return data.map(item => convertToType(Target, item));
   }
 
-  return Object.assign(new Target(), data);
+  return typeof Target.fromModel === "function"
+    ? Target.fromModel(data)
+    : Object.assign(new Target(), data);
 }
 
 export function getEnumValuesMap<T extends object>(enumObject: T) {


MichalLytek:
Nope, the mongoose-specific approach can't be hardcoded in the codebase.
All we can merge is some info in docs in examples.

> So there are couple of possible solutions how to avoid this:

3. Use the middleware which is presented in the examples:
https://github.com/MichalLytek/type-graphql/blob/master/examples/typegoose/typegoose-middleware.ts

syadykin:
@MichalLytek As I already mentioned in PR description soluton in example does not work for me because of subschema used. Also my solutions is *not* mongoose-specific, it just provides some helpful way for corner cases. Seems I need go with my own fork then.

MichalLytek:
You've mentioned only "type-graphql serializes User model" which is not calling `.toObject()` but just assign properties, thus lacking the document data.

The mentioned middleware does EXACTLY the same trick like your `fromModel` with `.toObject()` but just automatically when you return document from resolver.

You can write your own middleware which will call `.fromModel()` to serialize the document.

PR #806, opened 2021-03-06
Properly emit types nullability when `defaultValue` is provided 

### LINKED ISSUE #751: Inconsistent behavior for defaultValue and nullable on input types
**Describe the Bug**
A field in an input type or argument list is set to `nullable: true`, if a default is provided, even when the global nullable is false. Explicitly specifying `nullable: false` results in a definition error.

On the behavior regarding argument lists and input types with default values and nullable: https://github.com/graphql/graphql-spec/issues/570

I mentioned this in https://github.com/MichalLytek/type-graphql/issues/498#issuecomment-723673596, but created this as a separate issue since the linked issue also argues about object types (while this one is only about input types) and to increase visibility.

**To Reproduce**
```ts
@InputType()
class MyInput {
  // definition error, but should be `foo: Int! = 0`
  @Field(() => Int, { nullable: false; defaultValue: 0 })
  public foo!: number;

  // with global nullable: false results in `foo: Int = 0`, but should be `foo: Int! = 0`
  @Field(() => Int, { defaultValue: 0 })
  public foo!: number;
}
```

**Expected Behavior**

Lets a asume a string argument or input field:
- `nullable: true` (type-graphql behaves correctly)
  - omitted: resolver gets `undefined` (this is a "legacy" behavior of graphql-js)
  - passed `null`: resolver gets `null`
  - passed a string: resolver gets the string
- `nullable: false` (type-graphql behaves correctly) (this is the only case where a value **must** be passed when querying)
  - omitted: error
  - passed `null`: error
  - passed a string: resolver gets the string
- `nullable: true, defaultValue: null` (type-graphql behaves correctly)
  - omitted: resolver gets `null`
  - passed `null`: resolver gets `null`
  - passed a string: resolver gets the string
- `nullable: true, defaultValue: 'foo'` (type-graphql behaves correctly)
  - omitted: resolver gets string "foo"
  - passed `null`: resolver gets `null`
  - passed a string: resolver gets the string
- `nullable: false, defaultValue: null`: invalid definition (type-graphql behaves correctly)
- `nullable: false, defaultValue: 'foo'` (**type-graphql behaves incorrectly**: if `nullable: false` is set explicitly, then type-graphql incorrectly rejects the definition as invalid. if `nullable: false` is set implicitly through global setting, then type-graphql silently converts the field to a `nullable: true` field. both cases are incorrect behavior.)
  - omitted: resolver gets string "foo"
  - passed `null`: error
  - passed a string: resolver gets the string

MichalLytek:
Fixes #751


backbone87:
call me stupid, but `git clone` -> `npm ci` -> `npx gulp package` gives me type errors, i wanted to try this against my project, but i need to build it first

MichalLytek:
@backbone87 sorry, forgot to update imports in gulpfile after tsconfig change 😅 should be fixed now

backbone87:
seems fine to me, but i would bring up the point of having a legacy mode, since in my schema i have basically a lot of stuff which is now "actually" required. we use also apollo-codegen in the client and the input types basically allowed omission for these cases.

i have to think about this whole issue again.

MichalLytek:
This PR only changes `field: Type = value` into `field: Type! = value`, so now you can't pass null but you still can omit the arg/input field to retrieve the default value in resolver.

backbone87:
yes you are correct, i just wanted to verify that this is indeed what is happening in client side type gen

MichalLytek:
@backbone87 How's the verifying going? 😛

backbone87:
Oh, I already did this before my last comment. All good and working as expected

PR #827, opened 2021-03-21
Support for directives in emitSchemaDefinitionFile function

carlocorradini:
**Please, before doing anything about this pull request read my use case.**

## Use Case
Currently I'm working for a project that is based on apollo [Federation architecture](https://www.apollographql.com/docs/federation).
This implies the usage of custom directives.
In development all works without any issues and the experience is awesome (thanks!).
In production Apollo Server has the introspection property disabled and the Gateway is no longer responsible for fetching schemas from the services (directly) but using [Managed federation](https://www.apollographql.com/docs/federation/managed-federation/overview). To set up managed federation we must register all implementing service schemas providing a schema file (introspection query on the running service to fetch the schema is supported but I think is not so great expecially during CI/CD).
So we must have these schema definition files with Apollo Federation directives.
Moreover it can be really useful when using in combination with [GraphQL Inspector](https://graphql-inspector.com).

## emitSchemaDefinitionFile with directives
To solve *"the issue"* I've implemented the support of directives in the emitSchemaDefinitionFile function by simply passing ```directives: true``` option.
Emitting the schema with directives **does not use** the official `printSchema` from graphql-js but `printSchemaWithDirectives` from **@graphql-tools/utils** (https://www.graphql-tools.com/docs/api/modules/utils#printschemawithdirectives).
The *"only"* drawback (as far as I know) of using ```printschemawithdirectives``` is that it does not allow customization of printSchema options having to do with comments, so **`commentDescriptions` option is ignored**.

## Example
```typescript
import { emitSchemaDefinitionFile, buildSchema } from "type-graphql";

const schema = await buildSchema(...);

await emitSchemaDefinitionFile("./schema.graphql", schema, { directives: true });
```

In the future if ```graph-js``` will support directives in the *printSchema* function the ```directives``` option can be easily deprecated.
Or, if also *printSchema* require a flag this flag can be easily ported.

PS: This flag is false by default.

### DIFF
--- a/docs/directives.md
+++ b/docs/directives.md
@@ -67,6 +67,9 @@ class FooBarResolver {
 
 > Note that even as directives are a purely SDL thing, they won't appear in the generated schema definition file. Current implementation of directives in TypeGraphQL is using some crazy workarounds because [`graphql-js` doesn't support setting them by code](https://github.com/graphql/graphql-js/issues/1343) and the built-in `printSchema` utility omits the directives while printing.
 
+> If you want the directives to appear in the generated schema definition file you must set `directives` option to `true`. Be aware that emitting the schema with directives **does not use** the official `printSchema` from graphql-js but `printSchemaWithDirectives` from **@graphql-tools/utils** (https://www.graphql-tools.com/docs/api/modules/utils#printschemawithdirectives).
+> Moreover `printschemawithdirectives` it does not allow customization of printSchema options having to do with comments, so **`commentDescriptions` option is ignored**.
+
 Also please note that `@Directive` can only contain a single GraphQL directive name or declaration. If you need to have multiple directives declared, just place multiple decorators:
 
 ```typescript
--- a/docs/emit-schema.md
+++ b/docs/emit-schema.md
@@ -16,8 +16,9 @@ const schema = await buildSchema({
   // or pass a config object
   emitSchemaFile: {
     path: __dirname + "/schema.gql",
-    commentDescriptions: true,
+    commentDescriptions: true, // if 'directives' is true 'commentDescriptions' is ignored
     sortedSchema: false, // by default the printed schema is sorted alphabetically
+    directives: true, // emit the schema with directives
   },
 });
 ```
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
+        "@graphql-tools/utils": "^7.6.0",
         "@types/glob": "^7.1.3",
         "@types/node": "*",
         "@types/semver": "^7.3.4",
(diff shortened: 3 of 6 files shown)


MichalLytek:
`@graphql-tools/utils` is 10x heavier library than the whole `type-graphql`. It can't be a bundled dependency.

So unfortunately I have to reject your PR 😞 

What you could do is to make your own version of `emitSchemaDefinitionFile` that prints it with directives.
`buildSchema` + `printSchema` is all it does, so I don't see enough benefits of supporting this directly in `type-graphql` at a cost of 2MB dependency.

carlocorradini:
> `@graphql-tools/utils` is 10x heavier library than the whole `type-graphql`. It can't be a bundled dependency.
> 
> So unfortunately I have to reject your PR 😞
> 
> What you could do is to make your own version of `emitSchemaDefinitionFile` that prints it with directives.
> `buildSchema` + `printSchema` is all it does, so I don't see enough benefits of supporting this directly in `type-graphql` at a cost of 2MB dependency.

We can make it optional (like perrDeps) and if `directives` is true check if the library can be imported otherwise an error message is thrown. What do you think?

MichalLytek:
In NPM v7 peer deps are automatically installed, so it doesn't sound like a solution.

Doing lazy require-ing with errors about not installed also makes things unnecessary complicated, especially when the feature is so small. That's also why `class-validator` is still supported out-of-the-box even if it should be an optional feature hidden behind a plugin or something.

All we can do right now is to add a snippet for `emitSchemaDefinitionFileWithDirectives` in directives docs, below the warning about schema printing issues, so it can be easily found and adapted by user, as they might prefer to use `printSchema` from `@apollo/federation` instead of additional 2MB package.

carlocorradini:
> In NPM v7 peer deps are automatically installed, so it doesn't sound like a solution.
> 
> Doing lazy require-ing with errors about not installed also makes things unnecessary complicated, both for maintainers and the users themselves, especially when the feature is so small. That's also why `class-validator` is still supported out-of-the-box even if it should be an optional feature hidden behind a plugin or something.
> 
> All we can do right now is to add a snippet for `emitSchemaDefinitionFileWithDirectives` in directives docs, below the warning about schema printing issues, so it can be easily found and adapted by user, as they might prefer to use `printSchema` from `@apollo/federation` instead of additional 2MB package.

@MichalLytek 
I've updated the code with the dynamic import. It's only 5 lines of code 😊
It follows the same structure as TypeOrm for the drivers.
Let me know what you think 🤯

carlocorradini:
@MichalLytek 

I've updated the branch. Now the documentation is helping you with the new directives support in emitSchemaDefinitionFile function and also in the directives docs.

Now the function is a little bit shorter thanks to object destructoring assignement.

Please tell me what you think about it, dynamic import, docs etc...

PR #935, opened 2021-06-28
Support readonly arrays for roles of Authorized decorator

beaucollins:
The `@Authorized` decorator specifies `roles: T[]` or `...roles: T[]` as its argument. 

We have a constant sets of roles:

```ts
export const ADMIN_ROLES = ['a', 'b', 'c'] as const;
export const AdminRole = typeof ADMIN_ROLES[number];
```

We would like to use:

```ts
@Authorized<AdminRole>(ADMIN_ROLES);
```

TypeScript doesn't allow this because `readonly` arrays cannot be used in places where a non-`readonly` array is specified. However, a normal `Array` _can_ be used anywhere a `readonly Array<T>` is specified.

For now our workaround is spreading our constant array when used.

```ts
@Authorized(...ADMIN_ROLES);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - allow defining directives for interface types and theirs fields, with inheritance for object types fields (#744)
 - allow deprecating input fields and args (#794)
 - support disabling inferring default values (#793)
+- support readonly arrays for roles of `@Authorized` decorator (#935)
 ### Fixes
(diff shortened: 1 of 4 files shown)


MichalLytek:
@beaucollins Thank you for your contribution! ❤️ 

I've added test case, fixed lint error and added a changelog entry 😉

PR #950, opened 2021-07-14
Change to const to grab the `Context` and `Info`

timot7782:
Changed to const so that we can get the context and info if we need to

### DIFF
--- a/examples/apollo-federation/accounts/user-reference.ts
+++ b/examples/apollo-federation/accounts/user-reference.ts
@@ -1,6 +1,6 @@
 import { users } from "./data";
 import User from "./user";
 
-export const UserReferenceResolver = async (ref: Pick<User, 'Id'>, ctx: any, info: any): Promise<User> => {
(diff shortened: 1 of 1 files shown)


russelshane (approved):
👍👍

MichalLytek (changes requested):
But why?

timot7782:
> But why?

So that we can pass the context and info (fields) from apollo to the underlaying backend (if needed).

In our team, we need to grab the access token that is passed from the consumer to our microservices.

MichalLytek:
Yeah but:
1) You can do that with `function`, no need for `const`.
2) This is example and the example code is not using `context` nor `info`. Do you always declare unused parameters?
3) You can do that in your team, in your code, no need to introduce unused parameters in the examples. Have you referred to apollo federation docs about the `resolveReference` functions? Maybe we should create a docs page about integrating with `apollo-federation` instead?

So closing this PR as I don't see a reason for such code change 🔒 
Let's discuss your context and your issues, instead of your proposal of fixing the unknown root cause.

PR #1065, opened 2021-10-07
Add `browser` field to package.json for frontend use

campbellwmorgan:
* The `browser-shim` works nicely when trying to share code between
  frontend and backend, however manual overrides can be tricky
* Many build systems (including webpack) will take the script from
  the `browser` field simplifying this process


marcoklein (approved):

itpropro:
Any updates here @MichalLytek? I know its a small change, but it would help me fixing this in many projects and for the future.

MichalLytek:
I'm not convinced about this. 

TypeGraphQL can also normally work on frontend (e.g. https://github.com/MichalLytek/type-graphql/tree/master/examples/apollo-client). And looks like with this change, bundlers will start using shim instead of the real package 😕

ibilux:
> I'm not convinced about this.
> 
> TypeGraphQL can also normally work on frontend (e.g. https://github.com/MichalLytek/type-graphql/tree/master/examples/apollo-client). And looks like with this change, bundlers will start using shim instead of the real package confused

It does not work on framework like SvelteKit for frontend (but it does work for backend).

ibilux:
This PR fixed the problem in SvelteKit !

MichalLytek:
@campbellwmorgan 
I've updated your branch with recent `master` and updated the entry to match the new build folders introduced by #1400.

MichalLytek (approved):

PR #1111, opened 2021-11-12
docs(README): add note on how to download an example

omar-dulaimi:


MichalLytek (changes requested):
I don't think it's needed. People usually just clone the repo and run the examples because they need source code anyway.

omar-dulaimi:
I don't know about other people. But to me it's really helpful since the repo take a long time to download and might pose an issue for people with limited internet connections. It's used by Prisma examples as well.

PR #1120, opened 2021-11-19
feat(resolvers): add mapped types for resolver interface

jbmikk:
This adds the ability to map Apollo types to internal application types and have the resolver interface map the types accordingly for all field resolvers.

So a type mapped like this:

```
      @ObjectType()
      class SampleChildObject {
        [MappedType]: SurrogateObject;

        @Field()
        normalField: string;
      }
```
Makes the resolver interface get the root typed like this: 

```
      @Resolver(of => SampleChildObject)
      class SampleChildResolver implements ResolverInterface<SampleChildObject> {
        @FieldResolver()
        normalField(@Root() root: SurrogateObject) {
          return root.hiddenField;
        }
      }
```

This is backwards compatible, so a type with no mapping defined defaults to the original type.

### DIFF
--- a/src/interfaces/ResolverInterface.ts
+++ b/src/interfaces/ResolverInterface.ts
@@ -1,7 +1,15 @@
+export const MappedType = Symbol("MappedType");
+
+export type Mapped<T> = T extends { [MappedType]: infer M }
+  ? M
+  : T extends Array<{ [MappedType]: infer N }>
+  ? N[]
+  : T;
+
 /**
  * Resolver classes can implement this type
  * to provide a proper resolver method signatures for fields of T.
  */
 export type ResolverInterface<T extends object> = {
-  [P in keyof T]?: (root: T, ...args: any[]) => T[P] | Promise<T[P]>;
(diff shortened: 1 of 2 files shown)


MichalLytek:
Sorry, but your use case is too specific, and I don't want to bloat this library. 

Please use the custom resolver interface by yourself 😉

jbmikk:
I would argue to the contrary, the original `ResolverInterface` is too specific and Apollo is not really meant to be used like that. I just made it more generic to support more cases and still provided type safety, which I thought is what this library was about (because it's called `type-graphql`).
But anyway, I get your point about not bloating it. Maybe this should be just an article in the docs or an answer in SO.

MichalLytek:
> the original ResolverInterface is too specific and Apollo is not really meant to be used like that

It's not `apollo-typegraphql` library. I did plenty of projects with that without any issues. `ResolverInterface` was designed with TypeORM in mind, where your entity is your object type. It's just a fancy helper that guards against newbies mistakes 😛 You don't need to use that, and if you want some guard, write yours.

`[MappedType]: SurrogateObject;` is complicated and weird even for me, I can't imagine this being an official pattern in the docs.

PR #1122, opened 2021-11-21
add community generators section

capaj:


MichalLytek:
Honestly, I don't have a time to maintain a list of community plugins, generators and other packages.

I only list `typegraphql-prisma` because I maintain it. So I am sure it works well and is supported and safe to use.

If you found such info valuable, let's create a dedicated section for community packages, with a proper warning banner and list there all the framework and other tools related to TypeGraphQL, so that the community can maintain it via PRs.

PR #1141, opened 2021-12-15
add link to decorators proposal

capaj:


MichalLytek:
Do you think it will give the TC proposal more attention and push the work further? 😅

capaj:
The chance is slim, but I think it is good to have this anyway to show the users that this relies on non-standard javascript language feature. If it also raises the awarenes of this problem tiny little bit it is worth those 3 extra lines IMHO.

MichalLytek:
TBH I don't find that a big problem. Maybe in 2016 or so, when Angular 2 was born. Now I think it's safe to use it and from TypeGraphQL perspective, quite easy to migrate towards the TC39 JS solution.

I will close this PR because the Readme is already long enough 😉

PR #1143, opened 2021-12-15
feat(validation): added context to validation excecution

teovillanueva:
Hey! We are considering using type-graphql at Greener Bio. We need to validate some args with a value from the context and I've noticed you cant access the context in the `ValidatorFn`. This pull request adds the context to the `ValidatorFn`. It's a small PR so i think from the `CONTRIBUTING.md` it falls under the `Small Features` category. I've updated the tests accordingly and run the formatter, thats why there are so many modified files 😅


teovillanueva:
The best thing would be being able to run the auth-checker in an argument like this.

```ts
@GqlArgsType()
export class VehicleQueryArgs {
	@Authorized('ADMIN')
	@GqlField(() => ObjectIdScalar)
	vehicleId: ObjectId;
}
```

MichalLytek:
1) Remove Yarn.lock
2) Use npm and do not commit formatting changes. They make this PR hard to read.

teovillanueva:
Thanks for the feedback! I'll fix it.

teovillanueva:
Hey! Just pushed the changes. I found a better way to do it! Regarding using the auth-checker in the Arg or Args, where do you think i should start from?

teovillanueva:
I've just finished implementing the `@Authorized` for the `@ArgsType`  fields. Should i push it as a separate PR or can i include it here?

MichalLytek:
I had to fix some type issues in `ValidatorFn` so I reimplemented this feature in 2133c81 🔒 

PR #1174, opened 2022-01-18
chore: upgrade graphql dependency to major 16

### LINKED ISSUE #1100: GraphQL.js v16 support
Recently, v16 release of `graphql-js` has been published:
https://github.com/graphql/graphql-js/releases/tag/v16.0.0

It contains a lot of breaking changes, so TypeGraphQL needs to be upgraded in order to support it.

For now, please stick to the 15.x version as described in peer-depdencencies and what the runtime check says:
```
UnmetGraphQLPeerDependencyError: Looks like you use an incorrect version of the 'graphql' package: "16.0.0". Please ensure that you have installed a version that meets TypeGraphQL's requirement: "^15.3.0".
```

This thread will be updated with the investigation about that.
It won't happen soon, as all the other GraphQL ecosystem tools need to support `v16` too.

capaj:
fixes #1100 

All tests pass except this one
image

The only problem is that I cannot find the version of graphql-tools which had `SchemaDirectiveVisitor`, I have checked out the whole repo and greped it on older commits, like for example on `f0cb92e1361d10294b12a20188c2e1d57b92a543` and it's not there.
So if anyone can help me find source code for `SchemaDirectiveVisitor`, I can publish it on npm as separate module and we can fix that last test.


MichalLytek:
I appreciate your work but I already have almost working v16 upgrade on my own branch:
https://github.com/MichalLytek/type-graphql/tree/v16

I only haven't enough time to migrate directive tests to the new graphql tools because they now do not support transforming input values. I have to dig more or refactor the tests as I only care about AST, not about whole ecosystem flow.

PR #1242, opened 2022-03-24
WIP: V16 ESlint

tomasstrejcek:
just a tryout of migrating to eslint, regarding the rules, the migration would be pain in the a* so i would start rather with some minimalistic eslint config and gradually introduce new rules based on needs

for the current codebase to pass you need to disable tons of "recommended" rules

one of the biggest remaining hassles is tons of async functions without any await in them (as per @typescript-eslint/require-await) which I agree should be improved

one unnecessary complication are examples with react, maybe we should lint examples, tests and src separately?
(as per https://stackoverflow.com/questions/36762468/how-do-i-setup-a-folder-with-a-different-rule-and-another-folder-with-a-differen, never did it like that, need to try it)

@MichalLytek what is your opinion on standardjs? I use it in my projects all the time, love not having to deal with semicolons whatsoever :D

~~p.s.: maybe I didnt start on freshest v16 branch but that shouldnt really be an issue~~


tomasstrejcek:
btw moved @types to devDeps, they are not necessary for distributed packages, right? :)

tomasstrejcek:
@MichalLytek so, long time no feedback, what now? :D

PR #1264, opened 2022-04-13
docs(dependency-injection): improve example using TypeDI

th3oxen:
A refactor to give a clearer example on how to use TypeDI.


MichalLytek:

I agree that the docs require some upgrade and polishing, especially that there's no more container libraries for TypeScript.

However, your proposed changes are like embedding example repository inside the docs. We already have examples with source code, so the docs focus only on highlighting the most important parts and steps, not by embedding the full solutions like a copy-paste starter.

PR #1284, opened 2022-05-21
fix: use schemaName instead of methodName when building field resolve…

dobrynin:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

We are doing something more complicated, but I tried to boil it down to the essence of the issue below. Basically we have a mixin that adds very similar functionality that is parameterized by an argument `name`. We overwrite the `schemaName` appropriately by passing `name` to the field resolver, but when building the field resolver metadata, TypeGraphQL thinks that the subsequent field resolvers have already been added because it checks against the definition's `methodName`. I've run the test suite against the proposed change below, 

Simplified breaking code:
```
const withMixin = <TObjectType extends ClassType<{ id: string }>>(
  ObjectType: TObjectType,
  BaseClass: ClassType,
  name: string
) => {
  @Resolver(() => ObjectType)
  class FoobarResolver extends BaseClass {
    @FieldResolver(() => String, { name })
    async foobar(
      @Arg(`${name}Id`, () => ID) foobarId: string
    ): Promise<any> {
       ...Common logic we want to add via the mixin that is parameterized by `name`...
    }
  }
  return FoobarResolver;
};

### DIFF
--- a/src/metadata/metadata-storage.ts
+++ b/src/metadata/metadata-storage.ts
@@ -291,7 +291,9 @@ export class MetadataStorage {
           );
         }
 
-        const typeField = typeMetadata.fields!.find(fieldDef => fieldDef.name === def.methodName)!;
+        const typeField = typeMetadata.fields!.find(
+          fieldDef => fieldDef.schemaName === def.schemaName,
+        )!;
         if (!typeField) {
           const shouldCollectFieldMetadata =
             !options.resolvers ||
--- a/tests/functional/resolvers.ts
+++ b/tests/functional/resolvers.ts
@@ -2479,5 +2479,70 @@ describe("Resolvers", () => {
 
       expect(self).toBeInstanceOf(childResolver);
     });
+
+    it("should allow duplicate fieldResolver methods with different schema names for inherited resolvers", async () => {
+      getMetadataStorage().clear();
+      const INHERITED_DYNAMIC_FIELD_NAME_1 = "dynamicallyNamedMethod1";
+      const INHERITED_DYNAMIC_FIELD_NAME_2 = "dynamicallyNamedMethod2";
+
+      const withDynamicallyNamedFieldResolver = (
+        classType: ClassType,
+        BaseResolverClass: ClassType,
+        name: string,
+      ) => {
+        @Resolver(() => classType)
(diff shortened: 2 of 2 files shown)


MichalLytek:
@dobrynin Thanks for your contribution 🙌 
In order to merge your PR, we need you to create a failing test case that reproduce the issue and which is then fixed by your changes in the codebase.
This way any other future refactor won't affect your use case.

dobrynin:
@MichalLytek test added!

dobrynin:
Hi @MichalLytek , is it possible to get an estimate for when this change can be merged + released?

MichalLytek:
Please don't test metadata storage. It's implementation detail.
Your test cases should describe the real usage and real world assertion, so like "when I create this resolvers with this mixin, it should use this name, so that this field has this name and calls my field resolver properly".

dobrynin:
Okay @MichalLytek , I've refactored the test to stop relying on metadata storage and moved the test to the `resolvers.ts` file. Please take a look when you have a chance!

dobrynin:
Don't mean to nag you @MichalLytek, but please take a look at this one when you've had a chance. We'd like to stop using our fork of type-graphql :)

itpropro:
I have a similar scenario currently, any news when this is integrated into the main branch @MichalLytek ?

MichalLytek (approved):

MichalLytek:
@itpropro Sorry to keep you waiting so long for this simple fix, I constantly forgot to merge after rebase 😛 
Too much opened PRs and issues, too few time for all of that 😞 

@dobrynin Can you check if your branch or PR still have enabled "allow edits from maintainers" setting?
I can't push changes anymore and I wanted to fix the rebase issues in order to merge this PR finally 🙏

dobrynin:
@MichalLytek I don't see any option to do that, I think it may not be possible unfortunately because the fork lives under an organization: https://github.com/orgs/community/discussions/5634

MichalLytek:
@dobrynin So could you rebase your branch and fix the lint issue (missing `!` in test sample filed)? 🙏

dobrynin:
@MichalLytek done

MichalLytek:
@dobrynin Thank you very much for your contribution! ❤️

Issue #453, opened 2019-10-30
Types transformation utils


MichalLytek:
The upcoming Prisma 2 integration (#217) will generate TS classes with TypeGraphQL decorators based on data model definition of Prisma schema.

But to allow for further customization, TypeGraphQL needs some operators to transform the generated classes into the user one. Mainly to hide some fields from database like user password, so we need a way to omit some fields of the base class or pick only some selected fields.

Proposed API have two approaches to apply the transformation:
- pick/omit fields from emitted GraphQL schema using decorator option
```ts
@ObjectType({ pick: ["firstname", "lastname"] })
export class User extends BaseUser {
  // ...
}

@ObjectType({ omit: ["password", "salary"] })
export class User extends BaseUser {
  // ...
}
```
This approach is for use cases where you still want to have access to the hidden fields in other field resolvers, like hidding array or rates but exposing average rating by the field resolver.

- pick/omit fields both from emitted GraphQL schema and TS type using class wrapper
```ts
@ObjectType()
export class User extends Pick(BaseUser, ["firstname", "lastname"]) {
  // ...
}

@ObjectType()
export class User extends Omit(BaseUser, ["password", "salary"]) {
  // ...
}
```
This approach is better when you want to create a derived type, like a subset of some input type, so you won't accidentally use the not existing field.

Initial version might not support inheritance or have a prototype methods leaks, because it's not needed by the Prisma integration. In the next release cycle I will try to make it work with broader range of use cases.

Later, more types transformation utils will be implemented, like `Partial(Foo)` for making fields optional, `Required(Foo)` for making fields non-nullable.

If possible, maybe I will try to add even a mapping util that will map the keys of one type to new values in a new type. For example if you want to generate a sorting input (with field types only ASC/DESC) based on an object type which fields are representing database columns 😉

amille14:
> Later, more types transformation utils will be implemented, like Partial(Foo) for making fields optional, Required(Foo) for making fields non-nullable.

For anyone else looking for this functionality now, I was able to hack together this partial type factory function by directly modifying the metadata storage object. Not ideal as this is a private api but it seems to do the trick so far. It just copies fields of the given type and creates a new type with the same fields, only all nullable.

```ts
import { ClassType } from 'type-graphql'

// NOTE: This only works for object types, not input types

export function PartialType<E>(EntityClass: ClassType<E>): any {
  const metadata = (global as any).TypeGraphQLMetadataStorage
  class PartialClass {}
[...]

PR #1295, opened 2022-06-03
feat: Types transformation utils

xcfox:
This PR implements  #453 
Types transformation utils is a feature that has been on the back burner for a long time.

This PR comes with five utils types as: 
- PartialType: set all fields to nullable
- RequiredType: set all fields to NON_NULL
- PickType: only define specified field to a new type
- OmitType: omit specified field to a new type
- IntersectionType:  combines two types into one new type

Use it like so:
``` typescript
@InputType()
class PartialObject {
  @Field()
  nullableStringField: string;
}

@ArgsType()
class RequiredObject {
  @Field()
  nonNullStringField: string;
}

@InterfaceType()
class PickedObject {
  @Field()
  pickedStringField: string;
}

@ObjectType()
class OmittedObject {
  @Field()
  OmittedStringField: string;
}

@ObjectType()
class SampleObject extends IntersectionType(
  IntersectionType(PartialType(PartialObject), RequiredType(RequiredObject)),
  IntersectionType(
    PickType(PickedObject, "pickedStringField"),
    OmitType(OmittedObject, "OmittedStringField"),
  ),
) {}

```

It work with `class-validator` and `class-transformer`
Actually, this code has been in our own repository for a year, it refers to the implementation of nestjs and https://github.com/MichalLytek/type-graphql/issues/453#issuecomment-567049672

In addition, I didn't find a way to implement `mapping util` that ensures type safety.

### DIFF
--- a/src/helpers/inherit-metedata.ts
+++ b/src/helpers/inherit-metedata.ts
@@ -1,0 +1,186 @@
+// Inspired by @nestjs/mapped-types
+import { ClassType } from "../interfaces";
+
+export function applyIsOptionalDecorator(targetClass: Function, propertyKey: string) {
+  if (!isClassValidatorAvailable()) {
+    return;
+  }
+  const classValidator: typeof import("class-validator") = require("class-validator");
+  const decoratorFactory = classValidator.IsOptional();
+  decoratorFactory(targetClass.prototype, propertyKey);
+}
+
+export function inheritValidationMetadata(
+  parentClass: ClassType<any>,
+  targetClass: Function,
+  isPropertyInherited?: (key: string) => boolean,
+) {
+  if (!isClassValidatorAvailable()) {
+    return;
+  }
+  try {
+    const classValidator: typeof import("class-validator") = require("class-validator");
+    const metadataStorage: import("class-validator").MetadataStorage = (classValidator as any)
+      .getMetadataStorage
+      ? (classValidator as any).getMetadataStorage()
+      : classValidator.getFromContainer(classValidator.MetadataStorage);
+
+    const getTargetValidationMetadatasArgs = [parentClass, null!, false, false];
+    const targetMetadata: ReturnType<
+      typeof metadataStorage.getTargetValidationMetadatas
+    > = (metadataStorage.getTargetValidationMetadatas as Function)(
+      ...getTargetValidationMetadatasArgs,
+    );
+    targetMetadata
+      .filter(({ propertyName }) => !isPropertyInherited || isPropertyInherited(propertyName))
+      .map(value => {
+        const originalType = Reflect.getMetadata(
+          "design:type",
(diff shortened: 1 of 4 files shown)


MichalLytek:
Thank you for your PR ❤️ 

> It work with class-validator and class-transformer

I think `class-transformer` is not used by TypeGraphQL. Only NestJS is using that.

What I don't like about such monkey-patching is that it does not work with other decorator-based libraries like TypeORM or `joiful`. Having such official helpers with such issues would be really bad.

What do you think about you publishing this helpers as a npm package and then linking it in the docs here?
So that all users know about this but at any time can switch to an alternative implementation supporting other decorator-based libs or even own homecrafted or forked solution? 😉

xcfox:
@MichalLytek  Publishing this helpers as a npm package is a good idea. Do you think `type-graphql-transformation` is a good name?

itpropro:
I like the idea of having a separate npm module for transformation. Can you also update or close #453 @MichalLytek so that anyone who watches there for implementations of your initial proposal knows what the status is? Thank you!

ChrisLahaye:
Hi! We published our transformation utils to the [type-graphql-utils](https://www.npmjs.com/package/type-graphql-utils) package. It exports Pick, Required, Partial, and Omit. It doesn't have any specific code for class-validator like this PR, so I am not sure whether that works as we don't use it.

okaforcj:
hello @xcfox, @itpropro @MichalLytek  since this has not been published or merged in, i copied the helper functions source-code for transformation into my application and everything works like a charm, my only concern is that i upgraded to typegraphql 2.0 as well. Everything still works fine, i just wanted to know if there were any concerns to adopting this or any issue i'd face in the near future because i see that it uses metadata to understand most of whats going on, so i dont know there might be breaking changes or general concerns

okaforcj:
@xcfox  @ChrisLahaye i think i found a bug, if you use PartialType/Partial on arrays, it seems like it strips the array and give its parameriterised value i.e. when using partial type on array of strings, it will just return a string as metadata for graphql, so my graphql types are not reflecting the actual types anymore even though they'd have been annotated before passing it through the PartialType function, Any thoughts on this please?

ChrisLahaye:
@okaforcj Are you experiencing this issue on my library or the code here?

ChrisLahaye:
@okaforcj I see the same issue has been reported at https://github.com/ChrisLahaye/type-graphql-utils/issues/3. I will look into it.

ChrisLahaye:
@okaforcj I just released version `2.0.0` which should fix this issue. Credits to @caldarellial for reporting this issue and providing the fix.

okaforcj:
> @okaforcj I just released version `2.0.0` which should fix this issue. Credits to @caldarellial for reporting this issue and providing the fix.

that was quick! thanks for the help!

Issue #620, opened 2020-04-29
Custom Decorator to a class Resolver


muco-rolle:
**Is your feature request related to a problem? Please describe.**
It would be good to have a way to create a  resolver class decorator so that we would not have to repeat the decorator on each method of the resolver class

**Describe the solution you'd like**
Creation of a custom decorator 
```typescript

export function AuthGuard(): any {
    return createMethodDecorator(async ({ context }, next) => {
        // decorator business logic
    });
}
```
We should be able to decorate a class resolver
and every Query, Mutation in the resolver is executed after the decorator

```typescript
@Resolver()
@AuthGuard()
export class RecipeResolver {  
      @Query(() => [Recipe])
      recipes() {
             return [{ id: "1", name: "test" }];
       }

       @Query(() => Project)
        recipe() {
              return { id: 1, name: "test" };
        }
}
```

**Describe alternatives you've considered**

**Additional context**

Issue #200, opened 2018-11-19
Scoping middlewares


19majkel94:
Right now there are to ways to register a middleware:
- `@UseMiddleware` above the resolver method
- `globalMiddlewares` in `buildSchema`

In case of `@UseMiddleware` decorator, it should be possible to place in on the whole class to avoid manually placing on every method.

In case of global middlewares, it should be possible to register a global middleware that will be run:
- only once (e.g. dataloaders initialization)
- only in resolvers (query/mutation/subscription)
- only on advanced field resolvers (resolver's class method)
- only on simple field resolvers (getters, inline methods and simple fields)
and any combination of this possible options.
I have to investigate if the `@Middleware` decorator for class-based and `.scope` static property for function-based solution will be better than a detailed config object in `buildSchema`

In case of subscriptions, It should be also possible to choose if the middleware should be run only on subscribing, only on resolving the payload, or both.

This should also affect the `@Authorized` decorator which will be overwritten by a resolver's `@Authorized`, not concatenated like middlewares.

Yrobot:
So, if  i want to the log middleware just run once in one query, what should i do?

MichalLytek:
@Yrobot Use `@UseMiddleware` decorator to scope middleware to single query.
If by "one query" you mean one request, you need to use apollo server plugin or something similar.

Yrobot:
> use apollo server plugin

thanks, i use apollo server plugin in the end.

Yrobot:
> @Yrobot Use `@UseMiddleware` decorator to scope middleware to single query.
> If by "one query" you mean one request, you need to use apollo server plugin or something similar.

thanks, i use apollo server plugin in the end.

dallenbaldwin:
I came upon something there that is probably worth mentioning. Either it's a bug or the [documentation](https://typegraphql.com/docs/1.1.1/middlewares.html#attaching-middleware) is misleading.

I __cannot__ get a custom middleware to apply to an ObjectType Field when it has a Field Resolver in the ObjectType's Resolver. When attached to the Field Resolver, I get a callback for every returned _instance_ of that resolved field rather than once for the fact that the field was requested. Since this is a OneToMany resolution, I can't just get rid of the field resolver (without a lot of work)

I do think there's utility in having middleware run for every instance of a field when attached to a field resolver, however I think it would be more convenient to have middleware attached to a field run once if the field appears in a request, regardless of the return type

[...]

PR #1297, opened 2022-06-16
feat: Middleware Decorator to Resolver

xcfox:
This PR implements #620 and a part of #200  
This PR enhances `@UseMiddleware()` and `@Authorized()`, now we can put them directly on the resolver so that we can avoid repeating on each method. 

Use `@UseMiddleware()` like so:
``` typescript
@UseMiddleware(resolverMiddleware1)
@UseMiddleware(resolverMiddleware2, resolverMiddleware3)
@Resolver()
class SampleResolver {
  @Query()
  @UseMiddleware(middleware1)
  @UseMiddleware(middleware2)
  normalQuery(): boolean {
    middlewareLogs.push('normalQuery')
    return true
  }
}
```
Middlewares will be called in the order of global, resolver, field.

Use `@Authorized()` like so:
``` typescript
@Authorized()
@Resolver()
class SampleResolver {
  @Query()
  normalQuery(): boolean {
    return true
  }
  @Authorized('ADMIN')
  @Query()
  authQuery(): boolean {
    return true
  }
}
```
When you set `roles` for both the field and the resolver, the roles will be overwritten by the field's `@Authorized()`

I also make a function `createMiddlewareDecorator` to help you create custom middlewares, it is similar to `createMethodDecorator`, but can be placed on resolver


jbmikk:
@xcfox does this work with `@Directive` decorators too? That would be very nice.

MichalLytek:
@xcfox 
- I've renamed the custom middleware decorator functions
- added changelog entries
- updated docs to reflect the changes
- adjusted examples to use resolver class scoped middleware
- added missing shim export for this new decorator function

Now it should be ready to be merged!

MichalLytek (approved):

PR #1330, opened 2022-08-16
feat(createParamDecorator): custom param decorators can create Args

### LINKED ISSUE #1325: Custom Param Decorator breaks other @Arg if not last argument
**Describe the Bug**
We are using custom param decorators to inject args when derived by user, and allow overriding those args in other circumstances. For example, a normal user can only access their own account, while an admin user can access any account - so we use a nullable accountId argument and our decorator auto-fills it with the user's accountId if that user is a non-admin.

Here is some sample code:
```
export function AccountIdArg(
  argName: string = 'accountId',
): ParameterDecorator {
  // register a accountId param
  // override with current Account if not a staff user or not supplied
  const arg = Arg(argName, () => String, { nullable: true });
  const interceptor = createParamDecorator<OurCustomContextType>(({ args, context }) => {
    if (!context.currentUser) {
      throw new Error('User not logged in');
    }

    let accountId = args[argName];

    if (context.currentUser.userType === UserType.Consumer) {
      throw new Error('Not authorized');
    }

    if (context.currentUser.userType === UserType.Account) {
      accountId = context.currentUser.accountId;
    }

    if (!accountId) {
      throw new Error(`Missing required argument "${argName}"`);
    }

    return accountId;
  });
  return (target, key, index) => {
    interceptor(target, key, index);
    arg(target, key, index);
  };
}
```

The issue is, that when this custom param is the **only** argument, it works fine. If it's the **last** argument, it also works fine. But if it's somewhere in the middle (e.g. @Ctx() context, @AccountIdArg accountId, @Arg('input') input: SomeInputType), the third argument (input) will be undefined. Here are some screenshots of a debugger in different states:

**First Arg**

Notice how "input" is undefined
184040553-4680ea93-b0b5-485f-958e-29959766a891

**Second Arg**
184040475-9a1655c6-2098-4b5f-a8e4-a5577658ec66

Now "input" is working as expected - the only change is switching the order of the arguments.

**To Reproduce**
Use the decorator code above and set "currentUser.accountId" to some string in your context, then create a query or mutation  with the new param decorator.

**Expected Behavior**
The param decorator should work regardless of which order the argumetns are in.

**Logs**
See screenshot, debugger is clearest example

- OS: macOS & Amazon linux
- Node: 14.x
- Package version: 1.1.1
- TypeScript version: 4.7.4

russell-dot-js:
Fixes #1325

### DIFF
--- a/src/decorators/createParamDecorator.ts
+++ b/src/decorators/createParamDecorator.ts
@@ -1,20 +1,79 @@
 import { ResolverData } from "../interfaces";
 import { getMetadataStorage } from "../metadata/getMetadataStorage";
-import { SymbolKeysNotSupportedError } from "../errors";
+import { InvalidParamDecoratorError, SymbolKeysNotSupportedError } from "../errors";
+import { ArgOptions } from "./Arg";
+import { CustomParamOptions } from "../metadata/definitions";
+import { getParamInfo } from "../helpers/params";
+import { ReturnTypeFunc } from "./types";
+
+export type ParamOptions = {
+  arg?: {
+    name: string;
+    returnTypeFunc: ReturnTypeFunc;
+    options?: ArgOptions;
+  };
+};
+
+export type ParamResolver<TContextType = {}> = (resolverData: ResolverData<TContextType>) => any;
 
 export function createParamDecorator<TContextType = {}>(
-  resolver: (resolverData: ResolverData<TContextType>) => any,
+  resolver: ParamResolver<TContextType>,
+): ParameterDecorator;
+
+export function createParamDecorator<TContextType = {}>(
+  options: ParamOptions,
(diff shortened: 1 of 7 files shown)


russell-dot-js:
@MichalLytek coming up on 1 year old. Thoughts?

MichalLytek:
I like the idea but I see some tweaks needed in the implementation. Will take a look at that after #1400 👀

MichalLytek:
Thanks @russell-dot-js for your contribution! ❤️ 

Sorry that I took me so long to take care of this PR.
Unfortunately, because of conflicts made by #1400 and lack of required permissions to push the conflict resolution commit to your repo, I had to reimplement those changes with some tweak and additions in a new PR #1680, so I have to close this one 🔒

Issue #1397, opened 2022-12-13
Validation issues on version 2.0.0-beta.1 with @apollo-server@4.3.0


YMSpektor:
**Describe the Bug**
I'm using type-graphql v2.0.0-beta.1 together with @apollo-server@4.3.0. (As mentioned [here](https://github.com/MichalLytek/type-graphql/issues/1385) type-graphql v1.1.1 doesn't work fine with apollo-server v4) There are some issues related to validation:
1) I'm not able to call a mutation passing correct variables in the body, it leads to Argument Validation Error in the response. The issue seems to be related to "forbidUnknownValues" option in the class-validator library, at least setting it to false allows to avoid the error.

2) The"validationErrors" field is always missing in the ArgumentValidationError response (even if I set  forbidUnknownValues to false, add some validator like MaxLength and pass incorrect value - "validationErrors" will be missing in the response JSON)

**To Reproduce**
Here is the code:
```
import "reflect-metadata";
import { ApolloServer } from "@apollo/server";
import express from "express";
import { expressMiddleware } from "@apollo/server/express4"
import cors from "cors";
import http from "http";
import { ApolloServerPluginDrainHttpServer } from "@apollo/server/plugin/drainHttpServer";
import bodyParser from "body-parser";
import { buildSchema } from "type-graphql";
import { Arg, Mutation, Query, Resolver } from "type-graphql";
import { ObjectType, Field, ID, InputType } from "type-graphql";

@ObjectType()
export class User {
    @Field(() => ID)
    id: string;

    @Field()
    email: string;

    @Field({nullable: true})
    firstName?: string;

    @Field({nullable: true})
    lastName?: string;
}

@InputType()
export class AddUserInput {
    @Field()
    email: string;

    @Field({nullable: true})
    firstName?: string;

    @Field({nullable: true})
    lastName?: string;
}

@Resolver(User)
export class UserResolver {
    users: User[] = [];

    @Query(() => [User], { name: "users" })
    getAll(): User[] {
        return this.users;
    }

    @Mutation(() => User, { name: "addUser" })
    add(@Arg("user") userInput: AddUserInput): User {
        const user = new User();
        Object.assign(user, userInput);
        user.id = Math.floor(Math.random() * Number.MAX_SAFE_INTEGER).toString();
        this.users.push(user);
        return user;
    }
}

const bootstrap = async () => {
    const app = express();
    const httpServer = http.createServer(app);

    const schema = await buildSchema({
[...]

PR #1404, opened 2022-12-26
fix(validation): Fix arguments validation after updating the `class-validator`

kolpakov-p:
In order to get rid of [CVE-2019-18413](https://devhub.checkmarx.com/cve-details/CVE-2019-18413/) present in old versions of `class-validator`. There was a bug in the old version of `class-validator`, which lead to run validation with `forbidUnknownValues` enabled. So, to return the old behavior, enabling it explicitly now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
         "apollo-cache-control": "^0.11.6",
         "apollo-server": "^2.21.0",
         "apollo-server-plugin-response-cache": "^0.6.0",
-        "class-validator": "^0.13.1",
+        "class-validator": "^0.14.0",
         "del": "^6.0.0",
         "graphql": "^15.5.0",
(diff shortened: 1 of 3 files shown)


MedAmineFouzai:
Agree I had to set `validate:false ` in Apollo server configuration to avoid the unknown or undefined error that it throws after testing apis Wich completely removes the use of class-validator package . Thnx this has to be merged asap @kolpakov-p

MichalLytek:
@MedAmineFouzai You don't need to disable, you can set global options for type-graphql:
https://github.com/MichalLytek/type-graphql/issues/1397#issuecomment-1351432122

kolpakov-p:
I’m closing the PR due to the way to fix the issue is already exists ([mentioned by MichaLytek above](https://github.com/MichalLytek/type-graphql/pull/1404#issuecomment-1368516346)). Thank you!

Issue #1224, opened 2022-03-06
GitHub sponsors missing in the members sections


itpropro:
Currently only OpenCollective sponsors, not GitHub sponsors (https://github.com/TypeGraphQL) are updated in the README. As there are more sponsors on GitHub (at elast for member tier) and there is not much to see in regards to new code (as I understood, you currently develop the next version privately @MichalLytek), I think it would be a nice touch to have all the sponsors listed in the same place.

MichalLytek:
You're right, I was thinking about it but forgot at some time when some contributors on GitHub sponsors appeared.

Do you know if I can somehow automatically display the financial contributors via a generated image, like open collective is able to do?

itpropro:
There is a great toolkit for generating sponsor images by @antfu, which I have seen some people use (https://github.com/antfu/sponsorkit). I hope that helps as I don't have any experience in managing sponsors, only with sponsoring :)

itpropro:
I think it would be best if you as repository and org owner decide which format and highlighting of different levels you prefer for your sponsors, implement that and integrate it into some CI pipeline for all of your projects.
The options seem pretty straight forward to automate that in a very little time.

jahead:
@MichalLytek if no one is working on this, I think it would be a something I could play with on the weekend. #myweekendsarefun

MichalLytek:
To myself: integrate with this tool: https://github.com/antfu/sponsorkit

carlocorradini:
Working on PR #1400

MichalLytek:
@carlocorradini can we extract this into a separate PR? I would like to play with this in a quicker manner than the rest of this 1400 PR 😉

carlocorradini:
@MichalLytek Copy the changes made in another PR and revert them to avoid conflicts.

carlocorradini:
@MichalLytek 
I'm thinking about two options:
1. Single image with all sponsors
2. Multiple images (similar to what we have now) under the corresponding MD title/subtitle

MichalLytek:
@carlocorradini We need multiple images like we have now. We have different sponsor tiers, we can't have all together 😛 

Also sponsors should be clickable with link to ther website. So maybe we should use this tool only to generate image for gihub backers?

carlocorradini:
@MichalLytek It's possible if you check the example on the site.
Unfortunately I can't try to generate the images due to che API keys required :/

MichalLytek:
[...]

PR #1428, opened 2023-03-03
chore(sponsors): autogenerate GitHub sponsors image

MichalLytek:
Based on https://github.com/antfu/sponsorkit

Currently, only GitHub sponsors, as the support in `sponsorkit` for OpenCollective is limited.

We would also need a way to make those svg clickable, as when embedding as img, the linking capability are lost.

Closing #1224 🔒


carlocorradini:
Why the image is not clickable?

MichalLytek:
Because of modern web browser engine security restrictions? 🤔 

Clicks works only if you embed svg tags directly, like open the svg file in browser.
While using `img` tag (via markdown), it's just and image, not a dynamic svg element.

carlocorradini:
> Because of modern web browser engine security restrictions? 🤔
> 
> Clicks works only if you embed svg tags directly, like open the svg file in browser. While using `img` tag (via markdown), it's just and image, not a dynamic svg element.

😥

PR #1435, opened 2023-03-24
Use Date scalars from `graphql-scalars` instead of custom ones

MichalLytek:
- make `graphql-scalars` package a peer dependency
- use date scalars from `graphql-scalars`
- remove custom date scalars
- remove `dateScalarMode` option from `buildSchema`


carlocorradini:
Awesome 😎

PR #1463, opened 2023-06-11
chore: add lambda example

kang-heewon:
Add a simple labmda example.


carlocorradini:
@kang-heewon Nice! 🤗🥳

Can you update [`docs/examples.md`](https://github.com/MichalLytek/type-graphql/blob/master/docs/examples.md) and [`examples/README.md`](https://github.com/MichalLytek/type-graphql/blob/master/examples/README.md)

Thanks 👋

examples/aws-lambda/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  // build TypeGraphQL executable schema
+  cachedSchema ??= await buildSchema({
+    resolvers: [RecipeResolver],
+  });
+
+  // Create GraphQL server
+  cachedServer ??= new ApolloServer({ schema: cachedSchema });
+
+  return cachedServer.createHandler({})(event, context, callback);
+};
```
MichalLytek, line 21: This is the only difference in this `aws-lambda` example, compared to the base one.

I think it might be more valuable to create `Recipes` section in docs, where we can put some simple snippets like this.
I don't think that it's as complicated as typegoose or typeorm integration that it deserves own example.

MichalLytek (changes requested):

MichalLytek:
Implemented as snippets in docs in 615680e 🔒

PR #1484, opened 2023-08-11
chore(ci): support semver prerelease

carlocorradini:
Support prerelease semver in CI
RegEx is copied from the [official one](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string)


MichalLytek (approved):
Let's see 👀

PR #1610, opened 2024-01-13
chore(ci): fixed ci release and removed Node.js 16

carlocorradini:


.github/workflows/check.yml
```diff
@@ -133,8 +133,6 @@ jobs:
           npm run build:benchmarks
 
       - name: Check
-        # only run on newest node version
```
MichalLytek, line 136: Why removing this? Does it make sense to run this on all node verions?

.github/workflows/release.yml
```diff
@@ -17,7 +15,7 @@ jobs:
   release:
     name: Release package on NPM
     runs-on: ubuntu-latest
-    if: github.event.workflow_run.conclusion == 'success' && startsWith(github.event.workflow_run.head_branch, 'refs/tags/v')
+    if: github.event.workflow_run.conclusion == 'success' && github.ref_name == 'master' && startsWith(github.event.workflow_run.head_branch, 'v')
```
MichalLytek, line 18: What does it fix and how?

.github/workflows/check.yml
carlocorradini: No, it's better to run everything on all Node.js versions.
Who knows that something will break :/

.github/workflows/release.yml
carlocorradini: Run this workflow only if the previous workflow (check) is successful
AND
If running on master
AND
If the head_branch starts with v (tag that has already been checked in check)

MichalLytek (approved):
Let's see

PR #1637, opened 2024-02-21
docs(sample): add guide for integrating with azure functions

Mingyang-Li:
Hi, I noticed we don't have any guides on how to integrate this library with Azure Functions, so I wrote a [Medium article](https://mingyang-li.medium.com/serverless-graphql-how-to-integrate-azure-functions-with-apollo-server-and-typegraphql-969e838b4483) about it.

This PR involves adding a new page on the documentation website (`v2.0.0-beta.6`), plus a link to my article on how we can implement the integration, hopefully this will help more people

Update (2024-02-27)
1. Added a compact guide for integrating with azure functions
2. Added example project for No.1


carlocorradini:
@Mingyang-Li Can you add an example in the `examples` directory?
Thanks 👍🙏

MichalLytek:
I don't see any point of advertising medium article in the project docs.

It would make much more sense to copy/compact the article and add it as a proper doc page here (and maybe also add example as @carlocorradini said).

Doc section with link to article is just like typing "TypeGraphQL Azure Functions" in Google.

website/versioned_docs/version-2.0.0-beta.6/azure-functions.md
```diff
@@ -0,0 +1,116 @@
...
+import { ApolloServer } from '@apollo/server';
+import { startServerAndCreateHandler } from '@as-integrations/azure-functions';
+import { buildSchemaSync } from 'type-graphql';
+import { Container } from 'typedi';
+import { GraphQLFormattedError } from 'graphql';
+import { UserResolver } from 'YOUR_IMPORT_PATH'; // TypeGraphQL Resolver
+import { AccountResolver } from 'YOUR_IMPORT_PATH';  // TypeGraphQL Resolver
+
+// Bundle resolvers to build the schema
+const schema = buildSchemaSync({
```
MichalLytek, line 28: Is this necessary to build the schema sync? It's not recommended approach for TypeGraphQL

Mingyang-Li: Yes it is.

In apollo-server, the `schema` field is expecting `GraphQLSchema`, not the one wrapped in a promise.

```ts
import { ApolloServer } from '@apollo/server';

// using buildSchemaSync
  schema, // type: GraphQLSchema ✅
  introspection: true,
});

// using buildSchema
  schema, // type: Promise<GraphQLSchema> ❌
  introspection: true,
});
```

When I used `buildSchema`, I got this error:
```text
Type Promise<GraphQLSchema> is missing the following properties from type:
description
extensions
astNode
extensionASTNodes
and 20 more.
```

I see on type-graphql doc that it uses `buildSchema` for examples integrating with apollo-server. If I missed something, let me know.

MichalLytek: Yeah, you have to read about `await` and the top-level await support.
I bet Azure handlers can work with async functions too.

Mingyang-Li: I see why `buildSchema` can be beneficial in some cases.

But we've just tried doing this:
```ts
// indes.ts

let server: ApolloServer;

const startServer = async () => {
    resolvers: [
      EmployeeResolver,
      HistoryEmployeeResolver,
      CurrentStateRoleResolver,
      HistoryCurrentStateRoleResolver,
      FutureStateRoleResolver,
      HistoryFutureStateRoleResolver,
      ChangeProcessResolver,
      ClearinghouseResolver,
    ],

    emitSchemaFile:
      env.NODE_ENV === 'local' ? path.resolve('./src/schema.graphql') : false,
    container: Container,
    validate: true,
  });

  server = new ApolloServer({
    schema,
    introspection: true,
    formatError: (err: GraphQLFormattedError) => err,
  });
};

startServer().catch(console.error);

export default startServerAndCreateHandler(server, {
  context: async (args: AzureFunctionsContextFunctionArgument) => {
    return await transformContext(args);
  },
});
```
and we got this error: `Worker was unable to load function handler-graphql: 'Cannot read properties of undefined (reading 'startInBackgroundHandlingStartupErrorsByLoggingAndFailingAllRequests')'
[2024-03-10T21:37:31.043Z`

Seems like `startServerAndCreateHandler()` from `'@as-integrations/azure-functions` is unable to load schemas wrapped in a promise, I could be wrong.

Still, I won't see why we need to stick to using `buildSchema` just because our resolver methods use asyncronous operations. Are you able to setup an azure function without using `buildSchemaSync`?

MichalLytek: > Are you able to setup an azure function without using buildSchemaSync?

I don't use Azure but it's easily possible for AWS. Maybe you can use the same pattern.

> Still, I won't see why we need to stick to using buildSchema just because our resolver methods use asyncronous operations. Don't we all need to first build a proper gql schema before starting the server?

Sync build schema is just a workaround for envs that can't handle the promise version. Sync is always imitating and might not have all the stuff available in async version, so at least docs should show how to use them properly.

MichalLytek:
@Mingyang-Li 
I had to remove the example as it's only about config related to all Azure functions + index file, so docs should be sufficient for integrating TypeGraphQL by users.
The example was using own package.json + a build step, so it's not easy to run as our other examples.

MichalLytek (approved):

Issue #1321, opened 2022-08-03
Two mounted apollo servers (resolvers with inhertance) – second one lacks ArgsType in query


Alex0007:
**Describe the Bug**
We are creating two schemas, 2 initializing 2 apollo servers (`apollo-server-express`), then trying to mount them on two different paths: `/graphql` and `/cci`

**To Reproduce**
This example includes parts of code

```ts
  @ArgsType()
  export class PaginationArgs {
    @Field((type) => Int, { defaultValue: 0 })
    offset!: number;
  
    @Field((type) => Int)
    take!: number;
  }
  
  @ArgsType()
  export class OrderArgs {
    @Field({ defaultValue: "_id" })
    sortBy!: string;
  
    // @Field((type) => SortOrder, { defaultValue: SortOrder.desc })
    // order!: SortOrder;
  }

  @Resolver((of) => itemType, { isAbstract: false })
  class AbstractResolverClass {
    @Query(returns => Task)
    ['tasks'+Math.floor(Math.random() *1000)](
      @Args() { offset, take }: PaginationArgs,
      @Args() { sortBy }: OrderArgs,
    ) {
      
    }

    @FieldResolver()
    ['field'+Math.floor(Math.random() *1000)](
      @Args() { offset, take }: PaginationArgs,
      @Args() { sortBy }: OrderArgs,
    ): number {
      return Math.random()
    }
  }
  
  // example of build function
  export const buildCciApolloServer = async (httpServer: http.Server) => {
  const schema = await buildSchema({
    resolvers: [
      TaskResolver,
      // TestResolver
    ],
    nullableByDefault: true,
    scalarsMap: [{ type: Types.ObjectId, scalar: ID }],
    globalMiddlewares: [TypegooseMiddleware],
    authChecker: null
  });

  const server = new ApolloServer<ApolloContext>({
    schema,
    plugins: [
      ApolloServerPluginDrainHttpServer({ httpServer }),
      ComplexityPlugin({
        schema,
        maxComplexity: 100,
        estimators: [simpleEstimator({ defaultComplexity: 1 })],
      }),
    ],
    context: graphqlAuthContext,
  });

  return server;
};
  
  // init
  const cciApolloServer = await buildCciApolloServer(httpServer)
  const apiApolloServer = await buildApiApolloServer(httpServer)

  await cciApolloServer.start()
  await apiApolloServer.start()

  cciApolloServer.applyMiddleware({ app, path: '/cci' })
[...]

PR #1691, opened 2024-05-19
fix(metadata): restore original metadata to enable repeated buildSchema executions

syonthomas:
### Overview

This PR addresses the issue described in [#1321](https://github.com/MichalLytek/type-graphql/issues/1321), where the function [buildExtendedResolversMetadata](https://github.com/MichalLytek/type-graphql/blob/3a31dbe7818d4b0c5ce4733cc5e29da94855c8a8/src/metadata/metadata-storage.ts#L337) alters instance properties, causing multiple calls to buildSchema to exhibit unexpected behavior.

### Problem Statement

The primary problem is that buildExtendedResolversMetadata modifies the following instance properties of MetadataStorage:

- queries
- mutations
- subscriptions
- fieldResolvers

This mutation leads to issues when buildSchema is called multiple times, such as missing arguments in subsequent schema builds, as noted in the original issue.

### Solution

The proposed solution in this PR involves preserving the original state of MetadataStorage. On successive calls to buildSchema, the original instance of MetadataStorage is restored, ensuring that repeated schema builds do not encounter the unintended side effects of modified metadata.

### Implementation Details

- Save Original Metadata: Store the initial state of MetadataStorage upon the first invocation of buildSchema.
- Restore Metadata: Before subsequent calls to buildSchema, restore MetadataStorage to its original state.

### DIFF
--- a/src/metadata/metadata-storage.ts
+++ b/src/metadata/metadata-storage.ts
@@ -34,6 +34,8 @@ import {
 } from "./utils";
 
 export class MetadataStorage {
+  original: MetadataStorage | null = null;
+
   queries: ResolverMetadata[] = [];
 
   mutations: ResolverMetadata[] = [];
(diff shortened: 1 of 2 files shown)


syonthomas:
Further testing showed that this solution has a race condition and does not work for more than two buildSchema runs. I have another solution ready, but I will wait a couple of days to refresh the PR to see if this solution finally works.

MichalLytek:
@syonthomas Actually I've managed to fix this last week in a more reasonable fashion. 
Created a PR now, which supersede this PR 🔒 

https://github.com/MichalLytek/type-graphql/pull/1698

PR #1698, opened 2024-05-30
fix(schema): prevent issues with subsequent schema builds

### LINKED ISSUE #1321: Two mounted apollo servers (resolvers with inhertance) – second one lacks ArgsType in query

MichalLytek:
This PR is another approach (after #1691 by community) to fix an issue related to subsequent schema builds, especially when we have resolvers with inheritance and those resolvers queries/mutations contains args.

The approach to fix this was to just block subsequent metadata storage `.build()` actions, which were mutating the initial metadata storage values.

Fixes #1321 🚀

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 - make possible creating custom decorators on resolver class level - `createResolverClassMiddlewareDecorator`
 - support registering custom arg decorator via `createParameterDecorator` and its second argument `CustomParameterOptions` - `arg` (#1325)
 
+## Fixes
+
+- properly build multiple schemas with generic resolvers, args and field resolvers (#1321)
+
 ### Others
 
 - **Breaking Change**: update `graphql-scalars` peer dependency to `^1.23.0`
--- a/src/metadata/metadata-storage.ts
+++ b/src/metadata/metadata-storage.ts
@@ -80,8 +80,6 @@ export class MetadataStorage {
 
   params: ParamMetadata[] = [];
 
-  private hasAlreadyBeenBuilt = false;
-
   collectQueryHandlerMetadata(definition: ResolverMetadata) {
     this.queries.push(definition);
   }
@@ -176,12 +174,6 @@ export class MetadataStorage {
   }
 
   build(options: SchemaGeneratorOptions) {
-    if (this.hasAlreadyBeenBuilt) {
-      return;
-    }
-
-    this.hasAlreadyBeenBuilt = true;
-
     this.classDirectives.reverse();
     this.fieldDirectives.reverse();
     this.argumentDirectives.reverse();
@@ -222,11 +214,10 @@ export class MetadataStorage {
     this.argumentDirectives = [];
     this.classExtensions = [];
     this.fieldExtensions = [];
+
     this.resolverClasses = [];
     this.fields = [];
     this.params = [];
-
-    this.hasAlreadyBeenBuilt = false;
   }
 
   private buildClassMetadata(definitions: ClassMetadata[]) {
--- a/src/metadata/metadata-storage.ts
+++ b/src/metadata/metadata-storage.ts
@@ -80,6 +80,8 @@ export class MetadataStorage {
 
   params: ParamMetadata[] = [];
 
+  private hasAlreadyBeenBuilt = false;
+
   collectQueryHandlerMetadata(definition: ResolverMetadata) {
     this.queries.push(definition);
   }
@@ -174,6 +176,12 @@ export class MetadataStorage {
   }
 
   build(options: SchemaGeneratorOptions) {
+    if (this.hasAlreadyBeenBuilt) {
+      return;
+    }
+
+    this.hasAlreadyBeenBuilt = true;
+
     this.classDirectives.reverse();
     this.fieldDirectives.reverse();
     this.argumentDirectives.reverse();
@@ -214,10 +222,11 @@ export class MetadataStorage {
     this.argumentDirectives = [];
     this.classExtensions = [];
     this.fieldExtensions = [];
-
     this.resolverClasses = [];
     this.fields = [];
     this.params = [];
+
+    this.hasAlreadyBeenBuilt = false;
   }
 
   private buildClassMetadata(definitions: ClassMetadata[]) {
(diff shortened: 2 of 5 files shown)


syonthomas:
@MichalLytek , thank you for taking the time to address this issue and for your efforts with this PR. While it successfully resolves the associated test case, this solution does not work in my project.

I use buildSchema with different options. I suspect the issue arises because build now returns early even if other options are provided.

I will investigate further and try to reproduce the issue with a test case.

MichalLytek:
@syonthomas Thanks for the feedback. Could you share us the different options you use? Would be nice to add it to this PR as a failing test case.

syonthomas:
@MichalLytek You are welcome!

I've identified an issue with the handling of field resolvers. Below is a draft of a test case that will fail with the current implementation but will pass with my newest solution. In the second run `exampleFieldResolver` is missing in `OtherTestResolver`.

tests/functional/resolver.ts
 ```typescript
describe("Shared generic resolver", () => {
    beforeEach(async () => {
      getMetadataStorage().clear();
    });

    it("should handle field resolvers correctly on multiple buildSchema runs", async () => {
      @ObjectType()
      class TestResponse {
        @Field()
        data!: string;
      }

      @ArgsType()
      class TestArgs {
        @Field(() => Int, { defaultValue: 0 })
        testField!: number;
      }

      function makeResolverClass() {
        @Resolver(() => TestResponse)
        abstract class TestResolver {
          @Query(() => TestResponse)
          async exampleQuery(@Args() args: TestArgs): Promise<TestResponse> {
            return {
              data: `resolver ${args.testField}`,
            };
          }
        }

        return TestResolver;
      }

      @Resolver(() => TestResponse)
      class TestResolver extends makeResolverClass() {
        @FieldResolver(() => Boolean, { nullable: false })
        public async exampleFieldResolver(): Promise<boolean> {
          return true;
        }
      }

      @ObjectType()
      class OtherTestResponse {
        @Field()
        data!: string;
      }

      @ArgsType()
      class OtherTestArgs {
        @Field(() => Int, { defaultValue: 0 })
        testField!: number;
      }

      function makeOtherResolverClass() {
        @Resolver(() => OtherTestResponse)
        abstract class OtherTestResolver {
          @Query(() => OtherTestResponse)
          async exampleQuery(@Args() args: OtherTestArgs): Promise<OtherTestResponse> {
            return {
              data: `resolver ${args.testField}`,
            };
          }
        }

        return OtherTestResolver;
      }

      @Resolver(() => OtherTestResponse)
      class OtherTestResolver extends makeOtherResolverClass() {
        @FieldResolver(() => Boolean, { nullable: false })
        public async exampleFieldResolver(): Promise<boolean> {
          return true;
        }
      }

      const fistSchemaInfo = await getSchemaInfo({
        resolvers: [TestResolver],
      });

      const hasFoundFieldResolverInSchema = fistSchemaInfo.schemaIntrospection.types
        .some(type =>
          type.kind === 'OBJECT' &&
          type.name === 'TestResponse' &&
          type.fields?.some(field => field.name === 'exampleFieldResolver')
        );
      expect(hasFoundFieldResolverInSchema).toBeTruthy();

      const secondSchemaInfo = await getSchemaInfo({
        resolvers: [OtherTestResolver],
      });

      const hasFoundFieldResolverInOtherSchema = secondSchemaInfo.schemaIntrospection.types
        .some(type =>
          type.kind === 'OBJECT' &&
          type.name === 'OtherTestResponse' &&
          type.fields?.some(field => field.name === 'exampleFieldResolver')
        );
      expect(hasFoundFieldResolverInOtherSchema).toBeTruthy();
    });
  });
```

In my newest solution, I have changed schema-generator.ts to use a copy of `MetadataStorage`

```typescript
  static generateFromMetadata(options: SchemaGeneratorOptions): GraphQLSchema {
    this.metadataStorage = Object.assign(new MetadataStorage(), getMetadataStorage());
    this.metadataStorage.build(options);
```

Feel free to reach out if there's anything else I can assist you with.

MichalLytek:
@syonthomas 
Thanks for the feedback 👍 

I think this is the only solution for now. #183 could solve this but this is a major rewrite of the core, won't happen anytime soon.
For now let's just use your apporach. It's better than the first PR as we do not restore but make a local copy to work on.
Tests are green so I hope that we won't break anyones production setup 😃

PR #1767, opened 2024-11-12
AdonisJS integration

kerwanp:
Hi! We have recently released the first version of Adonis GraphQL, it is built on top of TypeGraphQL and I have to say that building this "adapter" with it has been a great journey!

This pull-request simply adds a short documentation page about Adonis GraphQL and how to get started.

Happy to discuss about changes, it might requires a disclaimer `Not maintained by TypeGraphQL`.

- Source code: https://github.com/FriendsOfAdonis/FriendsOfAdonis/tree/main/packages/graphql
- Documentation: https://friendsofadonis.com/docs/graphql

### DIFF
--- a/docs/adonisjs.md
+++ b/docs/adonisjs.md
@@ -1,0 +1,45 @@
+---
+title: AdonisJS Integration
+sidebar_label: AdonisJS
+---
+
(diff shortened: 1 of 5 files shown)


MichalLytek:
I think we should first introduce #1122 to gather all community-driven packages

MichalLytek:
I think that bullet list with like one to three sentences should be enough. There can be a link to npm or github or the website, so that users can navigate and read the full integration/package docs. No need to duplicate the detailed docs as they can quickly become outdated.

PR #1789, opened 2025-11-26
fix: prevent enumerable undefined properties in InputType instances (complete fix for #475)

### LINKED ISSUE #475: InputType returns undefined properties
**Describe the bug**
In latest 0.18.0-beta.5: If no value is passed to a nullable field on an InputType, an object is passed with that property undefined.

**To Reproduce**
```
@InputType()
export class MutationInput {
  @Field(() => String, { nullable: true })
  public nullable?: string;
}

@Mutation(() => MutationPayload)
public async mutation(
  @Arg('input') input: MutationInput,
): Promise<MutationPayload> {
  console.log(input) // { nullable: undefined }
}
```
`input === { nullable: undefined }`

**Expected behavior**
In latest stable version, the property is not returned:
`input === {}`

 - OS: MacOS
 - Node 11.13.0
 - Package version 0.18.0-beta.5
 - TypeScript version 3.5.3

**Additional context**
This makes TypeORM complain as it tries to set non-nullable fields to undefined.

AviBueno:
## Description

Fixes #475 (and completes the partial fix from commit 6f64ac4)

TypeGraphQL's `@InputType()` decorated classes were creating instances where ALL optional fields became enumerable properties set to `undefined`, even when not provided in the GraphQL query. This caused issues when using `Object.keys()`, `for...in`, or spread operators on input objects.

### Before This Fix

```typescript
@InputType()
class FilterInput {
  @Field(() => String)
  tenantId: string;

  @Field(() => String, { nullable: true })
  optionalField?: string;
}

// GraphQL Query: { filter: { tenantId: "123" } }

// In resolver:
Object.keys(filter)  // ['tenantId', 'optionalField'] ❌
filter.optionalField // undefined (but enumerable!)
```

### After This Fix

```typescript
Object.keys(filter)  // ['tenantId'] ✅
filter.optionalField // undefined (not enumerable)
```

## Root Cause

In `src/helpers/types.ts`, the `convertToType()` function was using:

```typescript
return Object.assign(new Target(), data);
```

Calling `new Target()` invokes the constructor, which initializes ALL `@Field()` decorated properties as enumerable with `undefined` values. `Object.assign()` only overwrites properties present in `data`, leaving unprovided optional fields as enumerable undefined values.

## The Solution

The fix now:
1. Calls the constructor to initialize instance fields (preserves existing behavior for non-GraphQL instance properties)
2. Removes undefined properties that weren't provided in the input data
3. Assigns the provided data via Object.assign()

```typescript
// Create instance by calling constructor to initialize instance fields
const instance = new (Target as any)();

// Remove undefined properties that weren't provided in the input data
// This prevents optional @Field() decorated properties from being enumerable
for (const key of Object.keys(instance)) {
  if (instance[key] === undefined && !(key in data)) {
    delete instance[key];
  }
}

return Object.assign(instance, data);
```

This creates an object with only the properties from the GraphQL input as enumerable, while preserving proper instance field initialization.

### Modified Files

- `src/helpers/types.ts` - Updated `convertToType()` function (added 9 lines)
- `tests/functional/inputtype-enumerable-properties.ts` - Added comprehensive tests

### Code Changes

```diff
 export function convertToType(Target: Function, data: any) {
   if (data == null) {
     return data;
   }
   if (Target instanceof GraphQLScalarType) {
     return data;
   }
   if (simpleTypes.includes(data.constructor)) {
     return data;
   }
   if (data instanceof Target) {
     return data;
   }
   if (Array.isArray(data)) {
     return data.map(item => convertToType(Target, item));
   }
-  return Object.assign(new Target(), data);
+  // Create instance by calling constructor to initialize instance fields
+  const instance = new (Target as any)();
+
+  // Remove undefined properties that weren't provided in the input data
+  // This prevents optional @Field() decorated properties from being enumerable
+  for (const key of Object.keys(instance)) {
+    if (instance[key] === undefined && !(key in data)) {
+      delete instance[key];
+    }
+  }
+
+  return Object.assign(instance, data);
 }
```

### New Test Cases

Added comprehensive test file `tests/functional/inputtype-enumerable-properties.ts` with 4 test cases:

1. **Basic optional fields** - Verifies Object.keys() only returns provided fields
2. **Nested InputTypes** - Ensures nested optional types aren't enumerable
3. **Provided optional fields** - Confirms provided fields ARE enumerable
4. **Explicit null values** - Verifies null (not undefined) is preserved

### Existing Tests

All existing tests pass (478 passed, 1 todo), confirming backward compatibility including tests that rely on instance field initialization.

## Real-World Impact

This bug affects production code that:

1. **Iterates over input properties:**
   ```typescript
   const fields = Object.keys(filter);
   for (const field of fields) {
     // Unexpectedly iterates over undefined properties
   }
   ```

2. **Uses property existence checks:**
   ```typescript
   if ('optionalField' in input) {
     // Incorrectly returns true even when not provided
   }
   ```

3. **Spreads input objects:**
   ```typescript
   const combined = { ...filter };
   // Includes undefined properties
   ```

4. **Deals with GraphQL client metadata:**
   - Apollo and other clients add `__typename`
   - Combined with enumerable undefined properties, causes validation errors

## Related Issues and History

- **Original Report:** #475 (November 2019) - "InputType returns undefined properties"
- **Partial Fix:** Commit 6f64ac4 (November 25, 2019) - Fixed argument conversion but not `convertToType()`
- **Incomplete Fix Noted:** March 23, 2022 - User @forsigner identified that `convertToType()` still had the issue
- **This PR:** Completes the fix by addressing the root cause

## Breaking Changes

**None.** This change makes the behavior more correct without breaking existing functionality:

- ✅ Property access (e.g., `input.optionalField`) works identically
- ✅ Type checking unchanged
- ✅ Validation behavior unchanged
- ✅ All existing tests pass (including instance field initialization)
- ✅ Only affects enumeration (Object.keys, for...in, spread)

## Verification

The fix has been tested with:
- ✅ Simple InputTypes with optional fields
- ✅ Nested InputTypes
- ✅ Array inputs
- ✅ GraphQL client metadata (`__typename`)
- ✅ Complex filter objects with multiple optional fields
- ✅ Instance field initialization (Math.random() fields in tests)
- ✅ All existing TypeGraphQL tests (478 passed)
- ✅ Production codebase via patch-package

### DIFF
--- a/src/helpers/types.ts
+++ b/src/helpers/types.ts
@@ -106,7 +106,18 @@ export function convertToType(Target: any, data?: object): object | undefined {
     return data.map(item => convertToType(Target, item));
   }
 
-  return Object.assign(new Target(), data);
+  // Create instance by calling constructor to initialize instance fields
+  const instance = new (Target as any)();
+
+  // Remove undefined properties that weren't provided in the input data
+  // This prevents optional @Field() decorated properties from being enumerable
+  for (const key of Object.keys(instance)) {
+    if (instance[key] === undefined && !(key in data)) {
+      delete instance[key];
+    }
+  }
+
+  return Object.assign(instance, data);
 }
 
 export function getEnumValuesMap<T extends object>(enumObject: T) {
--- a/tests/functional/inputtype-enumerable-properties.ts
+++ b/tests/functional/inputtype-enumerable-properties.ts
@@ -1,0 +1,168 @@
+import "reflect-metadata";
+import { type GraphQLSchema, graphql } from "graphql";
+import { Arg, Field, InputType, Query, Resolver, buildSchema } from "type-graphql";
+import { getMetadataStorage } from "@/metadata/getMetadataStorage";
+
+describe("InputType enumerable properties", () => {
+  let schema: GraphQLSchema;
+
+  beforeAll(async () => {
+    getMetadataStorage().clear();
+
+    @InputType()
+    class SampleInput {
+      @Field()
+      requiredField!: string;
+
+      @Field({ nullable: true })
+      optionalField?: string;
+
+      @Field({ nullable: true })
+      anotherOptional?: number;
+    }
+
+    @InputType()
+    class NestedInput {
+      @Field({ nullable: true })
+      optionalNested?: string;
+    }
+
+    @InputType()
+    class ParentInput {
+      @Field()
+      required!: string;
+
+      @Field(() => NestedInput, { nullable: true })
+      nested?: NestedInput;
+    }
+
+    @Resolver()
+    class SampleResolver {
+      @Query(() => String)
+      testSimpleInput(@Arg("input") input: SampleInput): string {
+        return JSON.stringify({
+          keys: Object.keys(input),
+          hasOptional: "optionalField" in input,
+          hasAnother: "anotherOptional" in input,
+          optionalValue: input.optionalField,
+        });
+      }
+
+      @Query(() => String)
+      testNestedInput(@Arg("input") input: ParentInput): string {
+        return JSON.stringify({
+          keys: Object.keys(input),
+          hasNested: "nested" in input,
+        });
+      }
+    }
+
+    schema = await buildSchema({
+      resolvers: [SampleResolver],
+      validate: false,
+    });
+  });
+
+  describe("optional fields not provided", () => {
+    it("should not create enumerable properties for undefined optional fields", async () => {
+      const query = `
+        query {
+          testSimpleInput(input: { requiredField: "test" })
+        }
+      `;
+
+      const result = await graphql({ schema, source: query });
+
+      expect(result.errors).toBeUndefined();
+      expect(result.data).toBeDefined();
+
+      const data = JSON.parse(result.data!.testSimpleInput as string);
+
+      // Only requiredField should be in Object.keys()
+      expect(data.keys).toEqual(["requiredField"]);
+
+      // Optional fields should not be enumerable
+      expect(data.hasOptional).toBe(false);
+      expect(data.hasAnother).toBe(false);
+
+      // But should still be accessible (undefined)
+      expect(data.optionalValue).toBeUndefined();
+    });
+
+    it("should handle nested InputTypes correctly", async () => {
+      const query = `
+        query {
+          testNestedInput(input: { required: "value" })
+        }
+      `;
+
+      const result = await graphql({ schema, source: query });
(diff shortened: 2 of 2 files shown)


AviBueno:
> I noticed that properties with undefined values may still appear in nested objects, causing: `{"a": 1, "b": undefined, "c": { "d": 2, "e": undefined } }` to end up as: `{"a": 1, "c": { "d": 2, "e": undefined } }`
> 
> I'm looking into integrating a fix for that, but it might incur an overhead.

**TL;DR**: False alarm - the fix works.

@MichalLytek after further investigation, I can confirm that my concern about nested objects retaining undefined properties was a false alarm (I mistakenly tested with the latest release code instead of testing with my fix).

The fix handles nested InputTypes correctly, and here's why:

The `convertToInput` function in `convert-args.ts` recursively processes nested InputTypes, but crucially, it calls `convertToType` at **every nesting level** (line 113). So for an object like `{args: {orderBy: [{ name: "asc" }]}`:

1. `convertToInput` processes the parent object
2. For the nested `orderBy` array elements, it recursively calls `convertToInput`
3. At each level, `convertToType(Target, filteredData)` is called
4. My fix cleans up constructor-initialized `undefined` properties **at each level**

I verified this by testing with a real-world scenario involving nested InputTypes - the fix works correctly throughout the object tree.

### Edge case (not a practical concern)

The only scenario where nested undefined properties could persist is if someone explicitly pre-instantiates nested objects in their InputType class definition:

```typescript
@InputType()
class Parent {
  child: ChildInput = new ChildInput(); // Anti-pattern
}
```

However, this is an anti-pattern for InputTypes since GraphQL inputs should come from the query, not defaults. This doesn't represent a real-world use case.

### Regarding the GitHub Code Scanning security warning

I noticed GitHub flagged a "Remote property injection" warning on the PR. When trying to view details, the link leads to a 404 (https://github.com/MichalLytek/type-graphql/security/code-scanning/11), and the repo doesn't seem to have a SECURITY.md configured (https://github.com/MichalLytek/type-graphql/security).

I believe this is a false positive - the code iterates over `Object.keys(instance)` (properties from the class constructor) and checks against `key in data` (where `data` comes from GraphQL's already-validated input). The property names are controlled by the class definition, not user input.

Please advise on how to proceed with this.

Thanks.

AviBueno:
@MichalLytek glad I could help. Thank you for taking this all the way to the finish line! 💪

PR #1812, opened 2026-06-08
perf: dispatch AuthMiddleware and middlewares synchronously when possible

pkeuter:
AuthMiddleware was always async, so awaiting a synchronous authChecker scheduled a microtask and allocated a promise for every @Authorized field. On queries that resolve many authorized fields this dominates CPU time (profiled at ~40k authorized field resolutions costing several seconds of event-loop time, almost all promise/microtask churn).

- AuthMiddleware: run the authChecker synchronously, only awaiting when the checker (or container instance) is actually promise-like.
- applyMiddlewares: synchronous-capable dispatch that only awaits when a middleware, container instance or resolver returns a promise. A chain of synchronous middlewares now allocates zero promises instead of one each.
- Widen MiddlewareFn/NextFn to MaybePromise<any> so synchronous middlewares are first-class.

Behavior is preserved: same grant/deny logic, the same result-or-nextResult resolution and the multiple-next() guard. Auth denial now throws synchronously, which GraphQL execution handles identically to a rejected promise.


MichalLytek (approved):
Could you also write some tests that somehow verifies that the new approach really works?

pkeuter:
@MichalLytek quick reply! Of course, (A)I've added the tests. I noticed that this change, on large queries (thousands of authorized calls) breaks the query time in half. As you can probably see, the change has been authored by AI, but I went through it and I saw no strange things. Additionally I've tested in our own codebase.

pkeuter:
@MichalLytek, I've added some more tests to increase the coverage. Would you be so kind to let the workflows run again so I can see if the coverage didn't drop this time? Thanks!

pkeuter:
@MichalLytek is there anything else you expect from me, or is this PR good as it is?

